PR #7, opened 2015-04-17
Support light and biome data

wtfaremyinitials:


wtfaremyinitials:
@roblabla @dcbartlett

Issue #13, opened 2015-08-30
Use prismarine-block


rom1504:
https://github.com/prismarinejs/prismarine-block

That way we have a common block API usable by everybody

rom1504:
done

PR #15, opened 2015-09-05
Use prismarine block

### LINKED ISSUE #14: Add doc
documenting the API of prismarine-chunk that users can use

rom1504:
Also add an api description in the readme and an example
Fix #14 and #13


rom1504:
can anybody look at this ? @roblabla @wtfaremyinitials ?

PR #21, opened 2016-05-30
start adapting for an hybrid pe pc implementation

rom1504:
* [x] example works
* [x] use height, width and length constants at more places
* [x] make pe implementation use only one buffer for consistency (almost there)
* [x] add setBiomeColor to pc api (make it do nothing)
* [x] make tests run on both pe and pc automatically
* [x] implement initialize for pe (can be a simple loop to begin with)
* [x] set a decent color by default in pe (for generator meant for pc) : no let's not do that, it would be really sub-optimal, we can think of the best way to integrate pc generators when it comes to that

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 (https://img.shields.io/badge/gitter-join%20chat-brightgreen.svg)]
 (https://gitter.im/PrismarineJS/prismarine-chunk)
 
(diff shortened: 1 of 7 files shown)


filfat:
Will prismarine-world also support mcpe in the future?

rom1504:
@filfat yeah that's the objective

PR #23, opened 2017-01-07
[Experimental] Implemented Chunk.Load for minecraft 1.9 protocol

Flynnn:


rom1504:
Nice !
Not the most efficient thing, but it has the advantage of existing and working.

The dump method doesn't do the expected thing though (it should return the chunk in the 1.9 format), so it might be better to just put a `throw new Error("not implemented")` for now.

I'll test it and merge if it works.

Thanks for working on this :)

rom1504:
Ok one problem with the current code : 
calling load 2 times with a different bitmap will result in overwriting the first load.

Changing the code to use section as 1.8 now does (see pc/1.8) will likely fix this.

I take a look at it later, but feel free to try if you understand what I mean ;)

src/pc/1.9/chunk.js
```diff
@@ -0,0 +1,479 @@
...
+        storage = storage << 1;
+        data = data >> 1;
+      }
+    }
+    return storage;
+  }
+
+
+  /*
+  eatPackedBlockLongs(rawBuffer, palette, bitsPerBlock) {
```
rom1504: can this commented code be deleted ? (we can keep the commit if there's useful info in that)

rom1504:
Seems to work, but will merge only when it's clear it works in fs and mineflayer ;)

rom1504:
working with mineflayer.
@Flynnn says it works with flying-squid, so I think that's good enough

rom1504:
@Flynnn the consistency test is failing, can you try to run it with `npm test` and see what is wrong ? (first git pull, I did some cleanup)

rom1504:
Nice, thanks for this work !

Flynnn:
running npm test illuminated very effectively what was wrong! Thanks for pointing it out

Issue #24, opened 2017-01-08
Have tests


rom1504:
That actually test something.

Simple way to do this : 
1. get example chunk for each chunk version
2. check all kind of data get/set at all kind of position

rom1504:
this is done by 819cd82352adbae99aeb9d98cf0949f2296eb6f0

Issue #19, opened 2016-02-29
Support both 1.8 and 1.9 chunk format


rom1504:
http://wiki.vg/SMP_Map_Format#Format

http://wiki.vg/Pre-release_protocol#Chunk_Section

https://github.com/PrismarineJS/node-minecraft-protocol/issues/231#issuecomment-186531996

This will help mineflayer and flying-squid update to cross-version (1.8 + 1.9)

rom1504:
this will be useful https://gist.github.com/Gjum/0375b643ec13a42ab3c0#file-chunk_-5_5-dump

rom1504:
see https://github.com/SpockBotMC/SpockBot/pull/218/files

Gjum:
C implementation https://github.com/JavaProphet/MinecraftClone/blob/245fca26f1911ff87c04ac774079acb60546e78a/MinecraftClone/src/ingame.c#L201

rom1504:
https://github.com/Steveice10/MCProtocolLib/blob/4ed72deb75f2acb0a81d641717b7b8074730f701/src/main/java/org/spacehq/mc/protocol/data/game/chunk/BlockStorage.java

rom1504:
@mhsjlw has started working on that at https://github.com/mhsjlw/prismarine-nbt

louwers:
Did the chunk format change in 1.10?

mhsjlw:
Nope, not yet :smile:. I doubt the chunk format will change again for a very, very long time as this is probably the most optimized chunk format we've seen (hence how complicated it is, especially in JavaScript).

louwers:
Here is another implementation in C: https://github.com/broese/mcbuild/blob/master/mcp_packet.c#L625

@mhsjlw This kind of stuff looks quite odd in Javascript, but why is it more difficult? Hard to get good performance?

mhsjlw:
No, JavaScript doesn't natively support longs which the 1.9 chunk relies on, we've tried ways around it that seem to be successful, but we're not anywhere close.

louwers:
What are still currently problems with your implementation?

rom1504:
http://lunarco.de/minecraft/chunks/

rom1504:
or https://download.rom1504.fr/minecraft/chunks/

rom1504:
WIP there https://gist.github.com/rom1504/2d6451dbea299099b861ac6acef6ef7d

polkovnikov-ph:
[...]

PR #25, opened 2017-01-08
implement 1.8 chunk bitmap (/section)

### LINKED ISSUE #8: Implement the 0x21 map_chunk bitmap
Minecraft network packet 0x21 map_chunk allows for a bitmap to specify which 16x16x16 minichunks are included and which are left out to be filled with air. Currently there is no way of using this, and the entirety of the chunk must be sent over the wire.

rom1504:
I believe it works but because of #24 it's not really thoroughly tested.
Fix #8 
Related to #19

### DIFF
--- a/example.js
+++ b/example.js
@@ -1,4 +1,4 @@
-var Chunk = require('./')("pe_1.0");
+var Chunk = require('./')("1.8");
 var Vec3 = require("vec3");
 
 var chunk=new Chunk();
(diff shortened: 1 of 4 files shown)


rom1504:
Indeed dump and load are actually mostly untested and don't seem to actually work.

Add some tests.

rom1504:
Ok so I believe it works now.
But I wouldn't say my test are really thorough.
For example I do not test with a bitmap != 0xFFFF

Although I believe it should work, I will test this with mineflayer before merging.

rom1504:
tested with mineflayer

rom1504:
testing with flying-squid before merging

polkovnikov-ph:
Holding fingers crossed for this PR.

rom1504:
okay flying-squid works with this.

PR #29, opened 2017-03-31
Add skyLightSent param to Chunk.load (pc/1.9)

iRath96:
This makes it possible to do
```
column.load(args.data,args.bitMap,args.skyLightSent);
```
instead of
```
column.load(args.data,args.bitMap);
```
in PrismarineJS/mineflayer, thereby adding support for dimensions other than the overworld.

I only implemented this for pc/1.9, but since the `skyLightSent` parameter is optional, this will not break backward compatibility and support for other versions can be added easily 🎉 .

### DIFF
--- a/src/pc/1.9/chunk.js
+++ b/src/pc/1.9/chunk.js
@@ -62,9 +62,37 @@ function loader(mcVersion) {
     }
   ]];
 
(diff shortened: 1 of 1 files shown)


mhsjlw:
@rom1504

PR #45, opened 2018-06-02
Switch to ProtoDefc

mhsjlw:
This is an experimental pull request for prismarine-chunk that switches from ProtoDef to ProtoDef*c*. This pull request is not close to ready, it requires many things to be done first, most of which are documented here: https://github.com/ProtoDef-io/node-protodefc/pull/7

As well, this means that this repository will only work for Node version >= 8.3 (which I have been assured is not an issue)

So, for some numbers:
- ProtoDefc
  - 7079/duration
- ProtoDef
  - 6273/duration

Yes. Your eyes are not deceiving you. ProtoDefc is slower, and I have tested it over 20 times. However, it is easily explained: ProtoDefc JIT compiles the protocol during the program execution. This certainly has a 1000+ms overhead. I predict that if this overhead is removed and ProtoDefc is 'warmed up', it will perform noticeably better on softwares like flying-squid. What's left:

- [ ] Actually make sure it generates a real world and it's not just doing useless computation (yes, I haven't even tested this yet)
- [ ] Test it against flying-squid and see if there is a noticeable improvement
- [ ] Benchmark them fairly, i.e.: loading the compiled JS from disk (perhaps consider just bundling this JS w/ a prismarine-chunk release and add that functionality to node-protodefc)
- [ ] Have node-protodefc ready for release

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "scripts": {
     "test": "mocha",
     "benchmark": "bench-csv ./benchmarks/chunk-1.9.js -o ./benchmarks/results/chunk-1.9.csv",
-    "lint":"standard",
-    "pretest":"npm run lint"
+    "lint": "standard",
+    "pretest": "npm run lint"
   },
   "repository": {
     "type": "git",
@@ -38,7 +38,6 @@
   },
   "dependencies": {
     "prismarine-block": "^1.0.0",
-    "protodef": "^1.3.1",
     "uint4": "^0.1.2",
     "vec3": "^0.1.3"
   },
--- a/src/pc/1.9/chunk.js
+++ b/src/pc/1.9/chunk.js
@@ -9,96 +9,116 @@ const X3_CHUNK_VOLUME = 3 * CHUNK_VOLUME
 
 const BUFFER_SIZE = (CHUNK_VOLUME * 3) + CHUNK_CROSS_SECTION
 
-const ProtoDef = require('protodef').ProtoDef
+const { ProtoDef, PartialReadError } = require('protodefc')
 const { readUInt4LE, writeUInt4LE } = require('uint4')
-const {reverseBits16, reverseBits32, reverseBits} = require('../../reverse-bits')
+const { reverseBits16, reverseBits32, reverseBits } = require('../../reverse-bits')
+const assert = require('assert')
 
 module.exports = loader
 
 function loader (mcVersion) {
   Block = require('prismarine-block')(mcVersion)
 
-  // MC counts the longs, protodef wants the bytes. This is responsible for that conversion.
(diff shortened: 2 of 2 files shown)


mhsjlw:
(TODO for me, format the code using standard and fix the `cycle_test.js`)

rom1504:
Can't you remove the compile step from the benchmark ?

rom1504:
If it is really slower there is little benefit...

rom1504:
But unless it's compiling wrong, it should not be slower.
Worse case compare to manually written code.

mhsjlw:
Numbers don't seem to do this pull request any justice, so I will be posting two comparing videos that use `native-voxel-worldgen`, one w/ ProtoDefc (this pull request) and one w/out.
ProtoDef:
https://transfer.sh/S0DWA/voxel_worldgen_test_protodef.mp4
ProtoDefc:
https://transfer.sh/ik2IN/voxel_worldgen_test_protodefc.mp4

src/pc/1.9/chunk.js
```diff
@@ -341,10 +361,10 @@ class Chunk {
         } = this.readSection(chunk.slice(offset), skyLightSent)
 
         offset += size
-        blocksAddition = this.eatPackedBlockLongs(value.dataArray, value.palette, value.bitsPerBlock)
-        blocklightsAddition = value.blockLight
+        blocksAddition = this.eatPackedBlockLongs(value.data_array, value.palette, value.bits_per_block)
+        blocklightsAddition = value.block_light
 
-        skylightsAddition = skyLightSent ? value.skyLight : Buffer.alloc(skyLightSize)
+        skylightsAddition = skyLightSent ? value.sky_light : Buffer.alloc(skyLightSize)
```
rom1504: snake_case is wrong

package.json
```diff
@@ -38,7 +38,6 @@
   },
   "dependencies": {
     "prismarine-block": "^1.0.0",
-    "protodef": "^1.3.1",
```
rom1504: well ok but shouldn't you add a dep to protodefc ?

src/pc/1.9/chunk.js
mhsjlw: ProtoDef compiler enforces it

package.json
mhsjlw: No, it's not published we need my ProtoDefc stuff merged and my node-protodefc stuff merged. Use npm link.

src/pc/1.9/chunk.js
rom1504: it should transform the names to camelCase then

rom1504:
new version of prismarine-chunk doesn't use protodef

PR #54, opened 2018-12-04
Proper lefele4 PR for 1.13

rom1504:


lefela4:
give me a second

src/pc/1.13/chunk.js
```diff
@@ -0,0 +1,454 @@
...
+  }
+
+  setBiomeColor (pos, r, g, b) {
+    // polyfill
+  }
+
+  getBlockType (pos) {
+    var cursor = getBlockCursor(pos)
+    let a = this.data.readUInt16LE(cursor)
+    if (a === 8575) a = 383 // Some fix with the chunk 23..
```
rom1504: Why is this needed ?
It seems unlikely this is the only case where it would happen.

lefela4: I am still working on something about that.

rom1504:
Not ready for merging for at least one reason : one byte cannot represent numbers above 255, and there are such blocks in 1.13.
This.data is represented as an 1.8-chunk-style buffer, which doesn't work here. I think we should instead use either : 
* the initial 1.13 buffer 
or 
* a normal js structure (an array or an object)

rom1504:
fixed that issue in https://github.com/PrismarineJS/prismarine-chunk/pull/55

still pb to fix somewhere because it doesn't read properly 383

maybe we should test with more chunks

PR #59, opened 2019-04-09
Added chunk handling for chunks without skyLight data in 1.8

skullteria:
This was necessary to load the chunks of the nether and end dimension
which have no skyLight data sent with them since the skyLight value in
those dimensions is 0.

Added proper chunk testing for chunks with Metadata value skyLightSent
false which signifies chunk data without skyLight data.

Resources:
[1.8 Chunk_Format](https://wiki.vg/index.php?title=Chunk_Format&oldid=7164)
[1.8 Chunk_Data packet which uses the 1.8 Chunk_Format](https://wiki.vg/index.php?title=Protocol&oldid=7018#Chunk_Data) 

This should solve following Issue: 
PrismarineJS/mineflayer#729

The way I implemented it was by storing a skyLightSent state attribute for the chunk itself if it has skylight data stored or not,
so that I can simply check that in a if condition and stop accessing and  writing something that doesn't exist or simply return 0.
This is a different approach to handle loading chunk data without skyLight data then the >=1.9 versions do. They simply fill in a 0 when the parameter skyLightSent is false, which is slower than my method.
**EDIT**
I pushed a second commit for reading and writing biome data at the right position out of chunk data buffer without skyLight data


src/pc/1.8/chunk.js
```diff
@@ -139,44 +140,47 @@ class Chunk {
...
+    this.skyLightSent = skyLightSent
+
     const SECTION_SIZE = Section.sectionSize(skyLightSent)
 
     const { chunkIncluded, chunkCount } = parseBitMap(bitMap)
+    console.log('cc: ' + chunkCount)
     let offset = 0
     let offsetLight = w * l * sectionCount * chunkCount * 2
-    let offsetSkyLight = w * l * sectionCount * chunkCount / 2 * 5
+    let offsetSkyLight = (this.skyLightSent) ? w * l * sectionCount * chunkCount / 2 * 5 : undefined
```
rom1504: why undefined ? that doesn't make sense

skullteria: Because at line 127 I reference offsetSkyLight when this.skyLightSent is true. So the assigned value of offsetSkyLight will only be undefined when this.skyLightSent is actually false, and that state won't change no matter what til line 179.
And because of that I choose undefined(rather than null which is still a obj) to not unnecessary calcute even once a offset for something I will never use. At first it looked like this to not even declare the variable to begin with 
`if (this.skyLightSent) let offsetSkyLight = (this.skyLightSent) ? w * l * sectionCount * chunkCount / 2 * 5`
and therefor be even faster. But since the variable would still needed to be referenced at line 179 even when only the condition is true I couldn't get around the declaration of offsetSkyLight so I thought atleast don't waste calculation time on an offset for a variable from which we already know that it won't get used but still must be declared to not cause errors.

rom1504:
Did you try that with mineflayer ? Does it fix the problem you were seeing ?

skullteria:
> Did you try that with mineflayer ? Does it fix the problem you were seeing ?

I ran my uncleaned version of this fix for weeks since I first told you guys on gitter that I was able to fix it.
There were no problems after testing it locally by going through the portal to the nether and back, walking far enough to load new chunks, logging in and out while being in the nether. In that way I was able to verify that both packets which are r receiving chunks were working correctly.
My productive bot which used the uncleaned version of the fix ran almost 24/7 on a heavily modded server(probably spigot) without any chunk loading problems. Every 15 min my bot was using home tp commands to teleport back and forth between the nether and the overworld.

rom1504:
Ok, great.
If there are some reusable parts in your bot, feel free to publish it as plugins, or even put your entire bot in GitHub, it can be interesting to other people ;)

derjp:
> > Did you try that with mineflayer ? Does it fix the problem you were seeing ?
> 
> I ran my uncleaned version of this fix for weeks since I first told you guys on gitter that I was able to fix it.
> There were no problems after testing it locally by going through the portal to the nether and back, walking far enough to load new chunks, logging in and out while being in the nether. In that way I was able to verify that both packets which are receiving chunks were working correctly.
> My productive bot which used the uncleaned version of the fix ran almost 24/7 on a heavily modded server(probably spigot) without any chunk loading problems. Every 15 min my bot was using home tp commands to teleport back and forth between the nether and the overworld.
> 
> **TL;DR**
> Yes, yes it fixxed the problem real good.

I ran my own bot after a long time to test this and wow, it really improved the stability. I didn't get any crashes in the 3-4 hours I tested it in. Before, it was much much worse.

Great work!

skullteria:
Thank you for the kind words. It makes me happy to read them :D

Issue #61, opened 2019-08-08
1.14 Support


Gaz492:
Are there any plans on adding 1.14 support?

rom1504:
yes but prerequisite it's #57  (since 1.13 and 1.14 have similar format)

PinkFlufflyLlama:
I'll be doing this, It could take a while because I have some other thing going on too  

> https://github.com/PrismarineJS/prismarine-chunk/issues/57#issuecomment-594141086

rom1504:
this could be done easily enough by adapting (/make a bit more generic) the new 1.13 implem

PinkFlufflyLlama:
Ok, I have much less time than I did before so I'm not going to be able to work on this anymore

PR #65, opened 2019-10-02
1.13 implementation

### LINKED ISSUE #57: reimplement 1.13 or 1.14 for real
use bit-buffer, it works well
serious example of use at https://github.com/MephisTools/diablo2-protocol/blob/master/lib/utils/itemParser.js

hornta:
Fixes #57, #61 

1.13 implementation

Added `Chunk.getMask` and removed parameter `bitmap` parameter from `Chunk.dump`.

I'm not able to run the benchmark, maybe because I'm on Win10 (https://paste.ee/p/hFZGX)

@rom1504 said in chat that it would be good to compare the results to quarry. I need some help with that.

- Removed package-lock.json from gitignore since it [should always be committed to respository](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5).
- Switched mocha for [Jest ](https://jestjs.io) instead of mocha to allow unit tests next to code and [delightful testing](https://github.com/jest-community/vscode-jest#the-aim).
- Dropped support for node 6, 8 and 10 to allow [bigint](https://nodejs.org/api/buffer.html#buffer_buf_readbigint64be_offset)
- Added myself as contributor (is it ok? :3)


.gitignore
```diff
@@ -1,5 +1,4 @@
 node_modules/
 dist/
 benchmarks/results/
-package-lock.json
```
rom1504: Please remove package-lock.json

README.md
```diff
@@ -15,21 +15,21 @@ A class to hold chunk data for Minecraft: PC 1.8, 1.9, 1.10, 1.11, 1.12 and 1.13
 ## Usage
 
 ```js
-const Chunk = require('prismarine-chunk')("1.8")
-const Vec3 = require("vec3")
+const Chunk = require("prismarine-chunk")("1.8");
```
rom1504: Incorrect linting, run standard --fix

src/pc/1.13/BitArray.test.js
```diff
@@ -0,0 +1,125 @@
+/* globals describe it BigInt */
```
rom1504: Why are you putting the test file in the same folder as source ?

src/pc/1.13/ChunkColumn.js
```diff
@@ -0,0 +1,270 @@
+const SmartBuffer = require('smart-buffer').SmartBuffer
+const Vec3 = require('vec3').Vec3
+const ChunkSection = require('./ChunkSection')
+const constants = require('./constants')
+const BitArray = require('./BitArray')
+const varInt = require('./varInt')
+const assert = require('assert')
+
+// wrap with func to provide version specific Block
+module.exports = (Block, mcData) => {
```
rom1504: Why version specific ? Does it work for any other version ?

rom1504:
The rationale for package-lock.json is no it should not be committed for packages that are used by other packages.
Indeed if there is a package-lock.json for these kind of packages the maintainers and the tests will use the version from the lock file while the dependents will use the version from package.json. The difference of version may introduce unseen bugs.

Package-lock.json is good for application, bad for libraries

rom1504:
Thank you for working on this !

src/pc/1.13/BitArray.test.js
hornta: I like having my tests closer to the source. I feel like it makes me not forget to add tests and it's clearer to see what things are missing tests. More of a personal opinion really but a dedicated test folder tends to get messy after a while. This is only concerning unit tests though.

src/pc/1.13/ChunkColumn.js
hornta: I doubt this ChunkColumn class works for other versions right now, that is why I send along the version specifics things I need to use here. maybe it could be generalized to tend for older versions but I don't feel like making it now.

.gitignore
hornta: yes will do. I've only ever developed applications so I listen to you 👍

hornta:
tests are indeed passing. I've teared my hair out these couple of late evenings just to understand the "palette" and how it should function.

rom1504:
what is the difference between chunk column and chunk section ?

rom1504:
setBlockData doesn't seem to be implemented ?

rom1504:
initialize is also not implemented

rom1504:
creating a simple chunk then dumping / loading it doesn't work

rom1504:
try out this branch https://github.com/PrismarineJS/mineflayer/pull/712
just run the test for it (`mocha -g "1.13"`), you'll see stuff that break

rom1504:
I guess we could have some basic tests based on node-minecraft-protocol + node-minecraft-wrap here

hornta:
> what is the difference between chunk column and chunk section ?

ChunkColumn(16x16x256) manages an array(16 in this case) of ChunkSections(16x16x16) and the biome data. 
By default the array is set to [null, null, null...] up to 16 which indicates that all sections is empty.

I didn't implement setBlockData because the current 1.13 impl. doesnt do it. 
Totally forgot to impl. the initialize with function method.

> creating a simple chunk then dumping / loading it doesn't work

Do you mean the internal tests(because they all passed as circleci reported) or the tests in mineflayer?

rom1504:
> ChunkColumn(16x16x256) manages an array(16 in this case) of ChunkSections(16x16x16) and the biome data.

yeah I understood that. Why not merge ChunkColumn and Chunk ?

rom1504:
> Do you mean the internal tests(because they all passed as circleci reported) or the tests in mineflayer?

I noticed that with mineflayer, but simply creating a chunk (new Chunk()) then doing a setBlockType and a dump/load will emit some errors. We could add a test for this

rom1504:
but I think trying to fix mineflayer test is a good way to see what needs to be fixed (the part of the test that complain about chunk anyway, for the rest I'll keep looking at it)

hornta:
I believe separating ChunkColumn and ChunkSection makes the code easier to reason about. Personally I try to avoid making unecessary large classes/functions if I really don't have to.

rom1504:
Yes I agree.
I was not proposing to merge ChunkColumn and ChunkSection but ChunkColumn and Chunk

hornta:
temporary pointed dep. to prismarine-block to my personal prismarine-block repo to continue local development of flying squid

Flynnn:
Hi, why is this not merged yet? Is there something left to be done?

rom1504:
I would like it if we could validate if it really works by making
mineflayer 1.13 work (see related issue on mineflayer side)

Flynnn:
Gotcha, that makes sense.

PR #66, opened 2019-10-08
fix dumping for noSkylight chunks on 1.9

IdanHo:
the 1.9 chunk implementation correctly differentiates between noSkylight chunks (like the nether and the end) and skylight chunks (the overworld) when loading chunk data, but it just assumes all chunks are skylight chunks when dumping. this pull request saves the type of the chunk when loading and dumps it correctly based on the saved type. (its possible the 1.13 implementation might have the same problem, but im less familiar with it)


IdanHo:
Might

IdanHo:
Accidentally closed it, but might want to add tests for it too

PR #75, opened 2020-04-17
Remove dependencie on vec3

Karang:
This removes every usage of Vec3 and replace it with anonymous objects.
Constructing objects is expensive.
This speeds a bit the internals of prismarine-chunks, but more importantly it relaxes the api by allowing the user to provide any object with x, y, z properties.
It doesn't break the compatibility with other modules that uses Vec3.

### DIFF
--- a/benchmarks/chunk-1.9.js
+++ b/benchmarks/chunk-1.9.js
@@ -1,7 +1,6 @@
 const blocks = require('minecraft-data')('1.12.2').blocks
 const chunk = require('../src/pc/1.9/chunk.js')
 const Block = require('prismarine-block')('1.12.2')
-const Vec3 = require('vec3')
 
 const blockCounts = Object.keys(blocks).length
 const Chunk = chunk('1.12.1')
@@ -17,8 +16,8 @@ function benchmark () {
     const c = new Chunk()
 
     c.initialize(initBlock)
-    const block = c.getBlock(new Vec3(0, 0, 0))
-    c.setBlock(new Vec3(0, 0, 0), block)
+    const block = c.getBlock({ x: 0, y: 0, z: 0 })
+    c.setBlock({ x: 0, y: 0, z: 0 }, block)
 
     const buffer = c.dump()
     const c2 = new Chunk()
--- a/package.json
+++ b/package.json
@@ -42,8 +42,7 @@
     "prismarine-block": "^1.2.0",
     "protodef": "^1.3.1",
     "smart-buffer": "^4.1.0",
-    "uint4": "^0.1.2",
-    "vec3": "^0.1.3"
+    "uint4": "^0.1.2"
   },
   "engines": {
     "node": ">=12"
--- a/src/pc/1.13/ChunkSection.test.js
+++ b/src/pc/1.13/ChunkSection.test.js
@@ -1,5 +1,4 @@
 /* globals describe test */
-const Vec3 = require('vec3').Vec3
 const ChunkSection = require('./ChunkSection')
 const constants = require('./constants')
 const assert = require('assert')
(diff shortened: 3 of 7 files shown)


rom1504:
It is more expensive to build an instance of Vec3 than an anonymous object containing the same properties ?
I find this highly surprising. V8 is supposed to optimize exactly this use case.

Do you have some benchmark to prove this ?

Are you sure this isn't caused by something that's wrong in Vec3 constructor / definition ?

Karang:
This is true for javascript objects in general.
You can have a look at this thread https://stackoverflow.com/questions/21545687/javascript-vs-new-object-performance which is a bit old, but provides a benchmark you can run.

Also, the goal here is to decouple from Vec3 so the users can use better implementations of vector if they want.

rom1504:
This thread is talking about in browser JavaScript that probably didn't
have the recent optimization of v8 and similar engines.
These new engines automatically convert properly defined objects to
structures in their internal compilation phase and hence make them much
faster than just hashmaps.

So if you're observing that generic objects are faster than a class-like
object like Vec3 there is something wrong with Vec3 that we can change.

If the objective is flexibility to use different kind of implementation of
Vec3, the user can always reimplement the functions we need from Vec3 (Vec3
then acting as an interface more than an implementation)

rom1504:
This page https://v8.dev/docs/turbofan has a bunch of link about what
optimization goes on inside v8

Karang:
I don't think forcing the user to implement the needed functions is a good idea, because the api itself forces bad optimization practices. On the other hand, if you look at the changes, you'll see that very little change was needed to make chunks vector generic (most of the changes are in the tests).

Here is an example of why the vec3 api is bad: https://github.com/PrismarineJS/prismarine-chunk/pull/75/files#diff-21aed50eb827e26a3fe4dac1b9e9fbfbL33

Here we want to take the modulo 16 of each component.
The vec3 modulus() takes a vec3-compatible object, so we need to create one.
```javascript
Vec3.prototype.modulus = function(other) {
  return new Vec3(
    euclideanMod(this.x, other.x),
    euclideanMod(this.y, other.y),
    euclideanMod(this.z, other.z));
};
```
Then it creates another Vec3 object to return. 
The euclideanMod function had to be made generic, so it cannot take advantage of the knowledge that the other vector components are power of 2:
```javascript
function euclideanMod(numerator, denominator) {
  var result = numerator % denominator;
  return result < 0 ? result + denominator : result;
}
```
This is heavier than just writing:
```javascript
{ x: pos.x & 15, y: pos.y & 15, z: pos.z & 15 }
```
Where we don't call any function, do 1 object construction instead of 2, and use a branchless modulo.

Another example is the constructor of Vec3: there is 2 possible way of creating the object, it all depends how you include it:
```javascript
const Vec3 = require('vec3') // heavy wrapper function (used in  src/pe/1.0/chunk.js)
const Vec3 = require('vec3').Vec3 // lightweight (used in test/test.js)
```
This can be missleading as most people don't read the code and will likely use the first one.

The vec3 implementation can be optimized quite a lot, I agree this should be done, but that is out of the scope of this PR. (If it was on github I could make a few suggestions)

If you want, I can put back the use of vec3 in the tests (as it is as good as any other library to provide the input), but I think the small changes in the chunk classes allow more flexibility.

rom1504:
Vec3 is on GitHub https://github.com/andrewrk/node-vec3 . I have rights on
it we can change it

I would rather we optimize vec3 than reimplement it inside prismarine-chunk.

The main issue with this PR is this custom x,y,z object that is rebuilt
every time which will most likely make things slow in some cases where v8
won't pick up it's the same structure.

About removing vec3 non constructor I think it would be good yes, I removed
all it usage from mineflayer already a while ago.

Do you have in mind any other implementation of Vec3 that we could use ?

rom1504:
https://v8.dev/blog/fast-properties

Looking at this, in this specific case it's probably the same in term of
performance to use Vec3 or not

rom1504:
So in this case we can merge this it's ok.

The main question/problem is a more general design issue.
What is the point of removing usage of Vec3 in prismarine-chunk if we'll still use it everywhere else ?

And if it's going to be used everywhere else we might as well fix it once and keep using it.

What do you think ?

Karang:
> And if it's going to be used everywhere else we might as well fix it once and keep using it.

ok

PR #77, opened 2020-04-27
Support for 1.14

### LINKED ISSUE #61: 1.14 Support

Karang:
Update of prismarine-chunk to 1.14 is blocked by the missing blocks.json in minecraft-data. See the issue: https://github.com/PrismarineJS/minecraft-data/issues/236.

Otherwise, this PR should provide basic support for 1.14.

In 1.14 the light system changed a lot, and light (sky and block) data are sent in separate packets, as well as having 2 more "light" sections per column (1 on top and 1 on bottom). For now I removed the light informations from the ChunkSection and plan to create a LightSection class to hold the light informations. It would nice to update minecraft-chunk-dumper to be able to dump the light update packets.

Also, 1.14 introduced the heightmap nbt in the chunk packet (not in the data) is it something we want to get an expose in the api ?

1.14 chunk format: https://wiki.vg/index.php?title=Chunk_Format&oldid=15025
1.14 Update light: https://wiki.vg/index.php?title=Protocol&oldid=15346#Update_Light

Closes #61


rom1504:
update mcdata to 2.47.0 and you'll get the missing data

rom1504:
> Also, 1.14 introduced the heightmap nbt in the chunk packet (not in the data) is it something we want to get an expose in the api ?

I think it's something that can be handled out of prismarine chunk

rom1504:
what do you want to do with light update packet ?

Karang:
> what do you want to do with light update packet ?

In 1.14, chunk update does not longer send light data. So we need light update data to initialize the chunk. I could be tempted to simply ignore the light data, as most of the bots will not use them, but in order to provide a complete implementation (for flying-squid or if someone need light for a bot) this needs to be handled properly.

> > Also, 1.14 introduced the heightmap nbt in the chunk packet (not in the data) is it something we want to get an expose in the api ?
> 
> I think it's something that can be handled out of prismarine chunk

ok, so we need to save the nbt in prismarine-chunk-dumper

rom1504:
ok so the API would then now need to have both a load and a loadLight functions. loadLight will do nothing for other versions

I think it makes for sense to at least for now not handle heightmap inside prismarine-chunk

rom1504:
I updated chunk dumper to also dump light data and used it for 1.14 and 1.15, you can rebase this PR to use that data
Would be great if you can finish this PR :)

Karang:
> I updated chunk dumper to also dump light data and used it for 1.14 and 1.15, you can rebase this PR to use that data
> Would be great if you can finish this PR :)

Thanks. I rebased and implemented the sky and block light sections load and dump.

rom1504:
released as 1.15

Issue #76, opened 2020-04-24
`minecraft-data` should be in dependencies


ianpaschal:
Right?

Now it's in `devDependencies` which _looks_ like a mistake to me. I'm happy to make PR as its only a 1 minute change but I wanted to confirm beforehand.

rom1504:
Yeah we used to only depend on prismarine-block so it didn't need to be a
dep. Now it seems we are using it directly so it should be a dependency
indeed

PR #81, opened 2020-05-09
1.15 chunks implementation

### LINKED ISSUE #80: 1.15 support
changes : 
* https://wiki.vg/index.php?title=Chunk_Format&type=revision&diff=15292&oldid=15271 biomes

I think that's it

So I think what we need to do is : 
* save biomes in dumper
* handle that in pchunk

Karang:
Copied 1.14 chunks and made the changes for getting/setting the biome in 1.15.
load(data) no longer read the biomes, dump() no longer write the biomes.
We need to add a loadBiomes and dumpBiomes() similar to loadLight() and dumpLight()

Closes #80 #76

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,22 @@ Returns the chunk raw data
 
 Load raw `data` into the chunk
 
+#### Chunk.dumpLight()
+
+Returns the chunk raw light data (starting from 1.14)
+
+#### Chunk.loadLight(data, skyLightMask, blockLightMask, emptySkyLightMask = 0, emptyBlockLightMask = 0)
+
+Load lights into the chunk (starting from 1.14)
+
+#### Chunk.dumpBiomes()
+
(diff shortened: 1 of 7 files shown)


rom1504:
No we don't need to add a separate method in this case since the data is coming from the chunk packet. We can simply optionally take the biomes in load method.
What do you think ?

Karang:
> No we don't need to add a separate method in this case since the data is coming from the chunk packet. We can simply optionally take the biomes in load method.
> What do you think ?

that works too

rom1504:
looking at the current API, I think you're right actually, it makes more sense to have a dumpBiomes (to avoid changing the return type of dump) and a loadBiomes
would be good to add that in the tests too

(I was a bit surprised that test passe but actually it makes sense since biomes is now out of the main chunk data)

Issue #43, opened 2018-05-21
support ground-up property


rom1504:
http://wiki.vg/Chunk_Format#Ground-up_continuous

needed for https://github.com/PrismarineJS/mineflayer/issues/666

BiPolarTeen:
I'm getting this issue whilst having my bot online. And a worldedit action is performed within the bots view. 
The error I'm getting is `Error: Data buffer not correct size (was 12288, expected 12544)`. Here is the chunk dump information
[dump.zip](https://github.com/PrismarineJS/prismarine-chunk/files/3095921/dump.zip)
 when `chunk.load` is called. 

This happens when worldedit sends chunk packets with the `groupUp` value being false.

rom1504:
done for 1.13, 1.14 and 1.15
for 1.8 and 1.9 -> 1.12 we could probably assume the same behavior (no biomes)

PR #86, opened 2020-06-11
Fix a json bug + biomes bug

Karang:
First bug:
When going to json and back again, the bitsPerValue property gets converted to string and never converted back. This causes an interesting issue at line https://github.com/Karang/prismarine-chunk/blob/045175e09f4e123f4ffa9b3c735dff6c74af0f67/src/pc/1.13/BitArray.js#L72 where the + become a concatenation...

Second bug:
As reported several time (last in date https://github.com/PrismarineJS/mineflayer/issues/1041), when fullChunk (groundUp) is false, the biomes array is not present in the packet. This PR adds the possibility to give a fullChunk parameter to the load function. Changes are needed in mineflayer to fix the bug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ Return the chunk bitmap 0b0000_0000_0000_0000(0x0000) means no chunks are set wh
 
 Returns the chunk raw data
 
-#### Chunk.load(data,bitmap=0xFFFF,skyLightSent = true)
+#### Chunk.load(data, bitmap = 0xFFFF, skyLightSent = true, fullChunk = true)
 
 Load raw `data` into the chunk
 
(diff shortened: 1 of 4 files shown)


rom1504:
here is the description about full chunk https://wiki.vg/Chunk_Format#Full_chunk
are we correctly handling the second behavior ? (overwriting stuff currently present in the chunk)

rom1504:
would be great to be able to close https://github.com/PrismarineJS/prismarine-chunk/issues/43

Karang:
> here is the description about full chunk https://wiki.vg/Chunk_Format#Full_chunk
> are we correctly handling the second behavior ? (overwriting stuff currently present in the chunk)

I think we can just call load() on the already existing chunk this will override only the sections present in the packet and keep the biomes from the chunk creation. It has to be tested when integrating with mineflayer.

> what about other versions ?

For 1.15, the biomes are not set in the load() function already. For <1.13, I couldn't find the proper documentation for the format.

PR #88, opened 2020-06-16
add overflow test for bitarray

rom1504:
I added this failing test to reproduce the problem (hit in https://github.com/PrismarineJS/flying-squid/issues/394)
I'm not sure how to fix. I thought you may have ideas on this @Karang 
issue : when assigning 8 bits to maximum value (15 in this case), the first int of the bitarray overflows and goes to negative. .set and .get works as expected with this. However writeBuffer doesn't work and I'm wondering about readBuffer
what do we want to do in this case ?
It should be possible to handle this without any negative int rights ?
I'm thinking we should do something in this line https://github.com/PrismarineJS/prismarine-chunk/blob/master/src/pc/common/BitArray.js#L69

### DIFF
--- a/src/pc/common/BitArray.js
+++ b/src/pc/common/BitArray.js
@@ -67,15 +67,15 @@ class BitArray {
 
     // Clear bits of this value first
     this.data[startLongIndex] =
-      (this.data[startLongIndex] & ~(this.valueMask << indexInStartLong)) |
-      ((value & this.valueMask) << indexInStartLong)
+      ((this.data[startLongIndex] & ~(this.valueMask << indexInStartLong)) |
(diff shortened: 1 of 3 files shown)


rom1504:
problem is caused by js bitwise operators (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators) operating on signed 32-bit numbers.
The code we have is assuming unsigned 32-bit numbers.
Simple illustration : 
```
> (15 << 28).toString(2)
'-10000000000000000000000000000'
> (15 << 27).toString(2)
'1111000000000000000000000000000'
> ((15 << 27)*2).toString(2)
'11110000000000000000000000000000'
```
We have to do something to take into account this

PR #90, opened 2020-06-25
SetBlockData for 1.13

Deudly:


Karang:
Could you also update :
* https://github.com/PrismarineJS/prismarine-chunk/blob/master/src/pc/1.15/ChunkColumn.js#L124
* https://github.com/PrismarineJS/prismarine-chunk/blob/master/src/pc/1.14/ChunkColumn.js#L126
With the same code ?
It would be one less thing to worry about later.

rom1504:
And the 1.9 one too please

Karang:
> And the 1.9 one too please

it is already implemented https://github.com/PrismarineJS/prismarine-chunk/blob/master/src/pc/1.9/ChunkColumn.js#L101

rom1504:
Ah yeah right it's a bit different so I did that already

Deudly:
Done!

PR #91, opened 2020-06-29
1.16

rom1504:
22:25 <Karang> let's say you have bitPerValue = 5, you can fit 12 x 5 bits values, and you'll have 4 bit remaining
      22:25 <Karang> in the old version, you would store the first 4 bit of the value in the same long, and the remaining 1 bit in the next
      22:25 <Karang> in 1.16, you store the 5 bits in the next, and you leave 4 bits of padding


rom1504:
just did plumbing
didn't adapt bitarray
that's the only thing needed for this to work (change get and set)

src/pc/common/BitArray.js
```diff
@@ -37,45 +39,92 @@ class BitArray {
...
-      ((this.data[startLongIndex] & ~(this.valueMask << indexInStartLong)) |
-      ((value & this.valueMask) << indexInStartLong)) >>> 0
-    const endBitOffset = indexInStartLong + this.bitsPerValue
-    if (endBitOffset > 32) {
-      // Value stretches across multiple longs
-      this.data[startLongIndex + 1] =
-        ((this.data[startLongIndex + 1] &
-          ~((1 << (endBitOffset - 32)) - 1)) |
-        (value >> (32 - indexInStartLong))) >>> 0
+    if (this.entriesSpanMultipleLongs) {
```
Karang, line 88: I think it would be better to have another BitArray.js, to keep the code clean and to avoid unnecessary branching in hot functions.

rom1504: you're right, I changed that

rom1504:
@Karang not passing tests, did you run these tests locally and had different results ?

rom1504:
ah I get it, you tried only bitarray tests

Karang:
> ah I get it, you tried only bitarray tests

yes

PR #102, opened 2020-10-20
Block entities

mat-1:
Allows `blockEntities` to be stored in the `ChunkColumn`s for each version
Related: PrismarineJS/flying-squid#451 & PrismarineJS/prismarine-provider-anvil#31


rom1504 (changes requested):
could you add something in the tests to make sure this work?

rom1504 (changes requested):
currently all data being set and get from chunks is going from methods in the API
I believe it would be cleaner to do the same here
* have appropriate get/set methods with the right positions
* put blockEntities in dump and load method

PR #104, opened 2020-10-22
Add Typescript Typings

### LINKED ISSUE #103: Add type hints.
Add type hints to enable better use of this lib and support with typescript

Darkflame72:
Will fix #103 and will aid in future development.


Darkflame72:
Currently it is all autogenerated so I will manually add the correct types so it isn't `any` for everything

rom1504:
Is there a way to avoid duplicating the typings for all the versions ? The interface followed should be exactly the same for any version

Darkflame72:
Yeah I should be able to do that. Are there any differences from an interface perspective for the dev using it per version or is it just internal?

rom1504:
No there should be no difference, we are using this without checking the version at all inside mineflayer (see https://github.com/PrismarineJS/mineflayer/blob/master/lib/plugins/blocks.js ) and inside prismarine-world.
The implementation per version is different due to mojang changing the network format.

That's one of the main point of prismarine-chunk actually, it hides away all the dirty implementation details of the network format behind an unique API.

Darkflame72:
Ok awesome I will have this done today and should be ready for review

Darkflame72:
This should fix that but I will double check everything to make sure and finish up types.

rom1504 (approved):

Issue #107, opened 2020-12-29
Investigate performance for generation


rom1504:
```js

const Chunk = require('prismarine-chunk')("1.16.4")
const { Vec3 } = require('vec3')

const generator = (x, y, z) => {
    if (y < 60) return 1
    return 0
}
const slowChunkGenerator = (chunkX, chunkZ) => {
    const chunk = new Chunk()
    for (let y = 0; y < 256; y++) {
        for (let x = 0; x < 16; x++) {
        for (let z = 0; z < 16; z++) {
            chunk.setBlockStateId(new Vec3(x, y, z), generator(chunkX * 16 + x, y, chunkZ * 16 + z))
        }
        }
    }
    return chunk
}

const fastChunkGenerator = (chunkX, chunkZ) => {
    const chunk = new Chunk()
    const p = new Vec3(0, 0, 0)
    for (p.y = 59; p.y < 60; p.y++) {
        for (p.x = 0; p.x < 16; p.x++) {
        for (p.z = 0; p.z < 16; p.z++) {
            chunk.setBlockStateId(p, 1)
        }
        }
    }
    return chunk
}

const {
  performance
} = require('perf_hooks');


const old = performance.now()
for (let i=0; i < 1000; i++) {
    slowChunkGenerator(0, 0)
}
console.log(performance.now() - old)
```

as part of the schematic dataset project, we noticed it's slow to generate chunks. It's particularly slow to generate full chunks.
This benchmark shows that it takes 1s when generating 1000 full chunk, and about 100x less for just one layer.

Also it seems in a browser environment (with webpack), it's much slower.

This issue is there to track this, I think at some point it might be interesting to check if we can improve this (check what happens in the browser, check if we can batch the set state ids, ...)

Karang:
When I run this benchmark I get 761ms for 1000 chunks (761us/chunk) for the slow version and 18ms for 1000 chunks (18us/chunk) for the fast version. How did you get 1s for a chunk ?

rom1504:
sorry typo, that's what I meant, about a second for 1000 chunks

rom1504:
by running the same benchmark in browser (to do that one way is to simply put the code above in https://github.com/PrismarineJS/prismarine-viewer/blob/master/examples/standalone/index.js (remove everything else there)) and `npm install && npm run prod-start` and open the browser

PR #108, opened 2020-12-29
optimize for browser by inlining getSectionIndex and removing asserts

rom1504:
we observed a speed up of x3 in firefox and x8 in chrome with these changes
with benchmark from #107

### DIFF
--- a/src/pc/common/BitArray.js
+++ b/src/pc/common/BitArray.js
@@ -1,4 +1,4 @@
-const assert = require('assert')
+// const assert = require('assert')
 const neededBits = require('./neededBits')
 
 class BitArray {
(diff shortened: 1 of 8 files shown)


src/pc/common/BitArray.js
```diff
@@ -1,13 +1,13 @@
-const assert = require('assert')
+// const assert = require('assert')
```
Karang, line 1: Shouldn't we remove it completely instead of just commenting ?

rom1504: I think these alerts have value, and ideally we would keep them but automatically remove them depending on environment.
Not sure exactly how to do that, but in the mean time I think it's useful to leave them commented so we have them here when we want to do it

Karang: ok

Karang (approved):

PR #110, opened 2021-03-07
Use Uint32Array instead of Array

Saiv46:


rom1504:
Can you post results of benchmarks showing it's faster ?

Saiv46:
> Can you post results of benchmarks showing it's faster ?
I wrote a little script for that (based on script used by Karang to compare JS and WASM version)

Result of benchmarking BitArray
```
js 330.3704000711441 (baseline)
js opt 243.9045000076294 (26%)
js bigint 33245.86610007286 (-9963%)
```

And NoSpan version:
``` 
js 589.248899936676 (baseline)
js opt 524.510999917984 (10%)
```

rom1504:
Using what benchmark script ?

Karang:
The fact that you reorganized everything makes it hard to review. Can you cleanup the diff, or list the changes ?

Saiv46:
> The fact that you reorganized everything makes it hard to review. Can you cleanup the diff, or list the changes ?
Sure

- In constructor:
- - Instead of `options` variables there's an object destructor
```diff
- constructor (options) # `if (options === null) return` <-- gay
+ constructor ({ bitsPerValue, capacity, data } = {})
```
- - `length` variable is inlined. Also, the formula is changed
```diff
- Math.ceil((options.capacity * options.bitsPerValue) / 64)
+ (this.capacity * this.bitsPerValue + 31) >> 5 # Inspired from WASM implementation
```
- - Array is replaced to `Uint32Array`
```diff
- Array(length * 2).fill(0) # https://v8.dev/blog/elements-kinds#avoid-creating-holes
+ new Uint32Array((this.capacity * this.bitsPerValue + 31) >> 5)
```
- - `this.{#} = ...` assigments are sorted alphabetically.
- In methods:
- - `neededBits` function in `resizeTo()` is inlined
- - ` | 0` and `Math.floor` are replaced to `>>> 0`, which is faster
- - Copy-pasted some methods from `BitArray` to `BitArrayNoSpan` and visa-versa (they're identical now)
- - **Much** more inlined variables in `get` and `set` methods

src/pc/common/BitArray.js
```diff
@@ -1,143 +1,73 @@
...
-    bitarray.capacity = parsed.capacity
-    bitarray.bitsPerValue = parsed.bitsPerValue
-    bitarray.valueMask = parsed.valueMask
-    return bitarray
+  constructor ({ bitsPerValue, capacity, data } = {}) {
+    this.bitsPerValue = bitsPerValue >>> 0
+    this.capacity = capacity >>> 0
+    this.data = data
+      ? (data.buffer ? new Uint32Array(data.buffer) : Uint32Array.from(data))
+      : new Uint32Array((this.capacity * this.bitsPerValue + 31) >>> 5)
```
Karang, line 7: This is wrong: if capacity and bitsPerValue are both odd, the length of the array will be odd too (which cannot be as its supposed to represent an array of Long (64bits)).

It is working for our current use case as we only use power of 2 capacity. But this may break in the future if mojang decide to do weird things (again). The computation we had before this PR was giving the correct result, and I very much doubt this change is bringing any performance improvement.

Saiv46: Okay, I'll revert it

rom1504:
> Using what benchmark script ?

rom1504:
do you intend to finish this?

Saiv46 (commented):
I reverted almost everything but that change as the title describes
And also now `fromArray` is both same in BitArray/BitArrayNoSpan

Saiv46:
@Karang I'm sorry for abandoning this PR for a long time.

This PR still retains performance gain, yet diff is much cleaner.

src/pc/common/BitArray.js
```diff
@@ -49,28 +42,13 @@ class BitArray {
...
-      i++
-    }
-    if (curBit > 0) {
-      data.push(curLong)
+    const bitarray = new BitArray({
+      capacity: array.length,
+      bitsPerValue
+    })
+    for (let i = 0; i < array.length; i++) {
+      bitarray.set(i, array[i])
```
Karang, line 50: You are missing the point of the fromArray function, the only reason it exist is to avoid looping and calling bitarray.set().
By inlining the set function, a few redundant masking operations can be removed, yielding a more efficient implementation.

Karang (changes requested):

src/pc/common/BitArray.js
Saiv46: Okay, then I don't get why BitArrayNoSpan::fromArray is implemented like that.

I'll revert that, but I still want to know

Karang: The reason is BitArrayNoSpan::fromArray was not done yet, it should be modified the same way.

rom1504 (approved):
lgtm

PR #111, opened 2021-03-08
Add cache to CI

Saiv46:


.github/workflows/ci.yml
```diff
@@ -21,5 +21,16 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
+    - name: node_modules cache
+      uses: actions/cache@v2
+      env:
+        cache-name: cache-node-modules
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
```
rom1504, line 30: we don't use package-lock.json ; how can this work ?

plus what's the benefit of this cache ? CI is already super fast (one minute)

PR #113, opened 2021-03-11
add blockentities

u9g:


rom1504:
would be great to finish

rom1504 (changes requested):
I see this pr does not have blockEntities in load/dump ; how are the block entities supposed to be used ?
could you also update the doc

rom1504:
let's redo it someday

PR #125, opened 2021-06-14
Add chunk implementation for 1.17

Archengius:
Removed skylight-related methods because skylight is no longer sent over network
Update biome and section implementations to support custom world height feature introduced in 1.17
Added data passed to chunk constructor to specify world's min/max height


src/pc/1.17/ChunkColumn.js
```diff
@@ -0,0 +1,266 @@
...
+const constants = require('../common/constants')
+const varInt = require('../common/varInt')
+const BitArray = require('../common/BitArrayNoSpan')
+
+// wrap with func to provide version specific Block
+module.exports = (Block, mcData) => {
+  return class ChunkColumn {
+    constructor (options) {
+      this.minWorldHeight = options.minWorldHeight ?? 0
+      this.maxWorldHeight = options.maxWorldHeight ?? constants.CHUNK_HEIGHT
```
extremeheat, line 12: Make sure these properties can be serialized to JSON and back

src/pc/1.17/ChunkColumn.js
```diff
@@ -0,0 +1,266 @@
...
+
+    setBlockType (pos, id) {
+      this.setBlockStateId(pos, mcData.blocks[id].minStateId)
+    }
+
+    setBlockData (pos, data) {
+      this.setBlockStateId(pos, mcData.blocksByStateId[this.getBlockStateId(pos)].minStateId + data)
+    }
+
+    // Skylight data has been removed in 1.17, now it is computed client-side automatically
```
rom1504, line 172: pchunk is used both server and client side
Even if it's not serialized, we still need to store it, if it's something that is needed for rendering

Archengius: I mean, realistically, we have no way of computing it without writing our own lighting engine. It's also never stored on server too, lightning is completely separate from chunk persistence and serialization now.
even though I could make it store lighting data, what would be the source of it?

Karang: There are ongoing projects to write a prismarine implementation of the lighting engine. And the world generators for flying-squid do generate (simple) light data. Independently of whether or not it is used somewhere, storing light is supported by other chunk version so it would be a breaking change to not support it here.

src/pc/1.17/ChunkColumn.js
```diff
@@ -0,0 +1,275 @@
...
+    getMask () {
+      return this.sectionMask
+    }
+
+    countVerticalBiomeGridSize () {
+      const worldTotalHeight = this.maxWorldHeight - this.minWorldHeight
+      return Math.floor((worldTotalHeight + 3) / 4)
+    }
+
+    countVerticalSections () {
```
Karang, line 71: This function calls a lot of tiny functions that could be inlined into:
```js
countVerticalSections () {
  return (this.maxWorldHeight - this.minWorldHeight) >> 4
}
```
Function calls are not free.

src/pc/1.17/ChunkColumn.js
```diff
@@ -0,0 +1,275 @@
...
+    setBlockType (pos, id) {
+      this.setBlockStateId(pos, mcData.blocks[id].minStateId)
+    }
+
+    setBlockData (pos, data) {
+      this.setBlockStateId(pos, mcData.blocksByStateId[this.getBlockStateId(pos)].minStateId + data)
+    }
+
+    // Skylight data has been removed in 1.17, now it is computed client-side automatically
+    // and never synced from the server in chunk packets
```
Karang, line 182: The light update packet still exists https://wiki.vg/Protocol#Update_Light so this claim should be double-checked.

src/pc/1.17/ChunkColumn.js
```diff
@@ -0,0 +1,275 @@
...
+const ChunkSection = require('./ChunkSection')
+const constants = require('../common/constants')
+const varInt = require('../common/varInt')
+const BitArray = require('../common/BitArrayNoSpan')
+
+// wrap with func to provide version specific Block
+module.exports = (Block, mcData) => {
+  return class ChunkColumn {
+    constructor (options) {
+      this.minWorldHeight = options.minWorldHeight ?? 0
```
Karang, line 11: If these properties are added to the api, they should be added to other versions as well with a constant value.

src/pc/1.17/ChunkColumn.js
Archengius: Alright, brought implementation back

src/pc/1.17/ChunkColumn.js
Archengius: That particular function is called exactly once per chunk creation and twice for chunk load, performance claim just makes no sense here.

src/pc/1.17/ChunkColumn.js
```diff
@@ -0,0 +1,275 @@
...
+      }
+      return chunk
+    }
+
+    // Utility methods used for retrieving section indices and checking block positions
+    getMask () {
+      return this.sectionMask
+    }
+
+    countVerticalBiomeGridSize () {
```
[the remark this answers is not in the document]
Archengius: Most of these functions are only used for implementation convenience in place of static functions taking relevant object fields as arguments, they are not supposed to be actually exposed to the API. Unfortunately JavaScript is a very convenient and flexible language, so access modifiers are not a thing here

src/pc/1.17/ChunkColumn.js
Archengius: fair point, brought lightning implementation back as requested before

src/pc/1.17/ChunkColumn.js
Archengius: But alright, I inlined it since that way it actually looks simpler

src/pc/1.17/ChunkColumn.js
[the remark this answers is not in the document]
Archengius: I cleaned it up maximally, leaving only functions for computing indices and amount of sections, hopefully that's enough

rom1504:
we need test chunk here, you can generate them with chunk dumper

rom1504:
https://github.com/PrismarineJS/minecraft-chunk-dumper

rom1504:
Tests failing, can you update the node version in ci file ?

Archengius:
Updated the version and chunk files to use new lighting packet format, as discussed in chunk-dumper PR

src/pc/common/BitSet.js
```diff
@@ -0,0 +1,104 @@
...
+const ADDRESS_BITS_PER_WORD = 5
+
+// Returns index of the word containing given bit
+function getWordIndex (bitIndex) {
+  return bitIndex >> ADDRESS_BITS_PER_WORD
+}
+
+// implementation of the bit set of variable length
+// heavily based on Java BitSet class
+class BitSet {
```
Karang, line 11: This class is not needed, use BitArray instead, there is no need for duplicated code.

src/pc/1.17/ChunkColumn.js
```diff
@@ -0,0 +1,474 @@
...
+
+      chunk.minWorldHeight = parsed.minWorldHeight
+      chunk.maxWorldHeight = parsed.maxWorldHeight
+
+      return chunk
+    }
+
+    // Utility methods used for retrieving section indices and checking block positions
+    getMask () {
+      return this.sectionMask.getWordAt(0)
```
Karang, line 79: What about just returning sectionMask and let the user deal with it ? I don't think its used a lot anyway, mineflayer use it only in internalTests.js, the usual preferred method is to get the section array and check non-null sections.

src/pc/1.17/ChunkColumn.js
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    getMaskArray () {
```
Karang, line 82: I don't think this method is needed at all, the mask should be returned integrally in getMask()

src/pc/1.17/ChunkColumn.js
```diff
@@ -0,0 +1,474 @@
...
+          section.writeBuffer(smartBuffer)
+        }
+      })
+
+      return smartBuffer.toBuffer()
+    }
+  }
+}
+
+function makeBitSetFromBitMap (bitMap) {
```
Karang, line 450: not needed if using bitArray

Karang (commented):
some minor comments, otherwise LGTM

rom1504:
@u9g could you please rebase this on top of master (and the chunks you added) and maybe make the missing changes ?

rom1504:
This needs to be adapted to:
* Use the raw bitmaps
* Use existing bit array class instead of bitset
* Make the tests pass

Would be great if anyone that know these things well could have a look (maybe @Karang or @extremeheat )

extremeheat:
Looks like the tests are failing because the light data is stored in one file now instead of two. BitSet seems functionally the same as BitArray with 1 BPV, not sure why the need for the extra class @Archengius.

extremeheat:
The data for the tests is also missing the section bit masks, renamed to "primaryBitMask" in 1.17

rom1504:
Yeah, those changes would need to be done by someone. I think Archengius has left this for other people to finish

Bobby99as (approved):

Bobby99as (approved):
it seems good

rom1504:
no, the CI is not passing yet

test/test.js
```diff
@@ -20,6 +20,17 @@ const depsByVersion = versions.map((version) => {
...
+  const isPostFlattening = version.startsWith('1.13') || version.startsWith('1.14') ||
+    version.startsWith('1.15') || version.startsWith('1.16') || version.startsWith('1.17')
+
+  const serializesLightingDataSeparately = version.startsWith('1.14') || version.startsWith('1.15') ||
+    version.startsWith('1.16') || version.startsWith('1.17')
+
+  const newLightingDataFormat = version.startsWith('1.17')
+
+  const serializesBiomesSeparately = version.startsWith('1.15') || version.startsWith('1.16') ||
+    version.startsWith('1.17')
```
Eagle-Anne, line 32: Should these be inverted wherever used and renamed e.g. to isPreFlattening? The pre-flattening versions will always remain the same so that won't need edits, but every time there's a new release it will be a post-flattening version and this will need to be updated or else risk buggy behavior.
It might make more sense to special case the "old behavior" on all of these: isPreFlattening, doesNotSerializeLightingDataSeparately, pre1dot17lightingDataFormat, doesNotSerializeBiomesSeparately, etc.

Eagle-Anne (changes requested):

Kashalls:
@Archengius Do you need any help completing this?

u9g:
> @Archengius Do you need any help completing this?

He hasn't worked on it for weeks, someone needs to finish it.

Kashalls:
> > @Archengius Do you need any help completing this?
> 
> He hasn't worked on it for weeks, someone needs to finish it.

I'm not sure what needs to be done.

rom1504:
I rebased on top of master

rom1504:
I fixed a few things.
Remain to be done here:
* remove bitset file and use the one we already have
* undo any breaking change to the API
* make cycle test pass: this needs investigation of the buffer before/after

rom1504:
useful here:
* https://wiki.vg/Chunk_Format 1.16 format

@extremeheat I think we had some info about 1.16 -> 1.17 diff, can you share it here ?

extremeheat:
Diff for 1.16 -> 1.17 : https://github.com/phase/1.16-to-1.17

rom1504:
done in https://github.com/PrismarineJS/prismarine-chunk/pull/133

PR #127, opened 2021-07-05
add version property to chunk object

u9g:


rom1504 (changes requested):
what do you think about returning the actually used version https://github.com/PrismarineJS/node-minecraft-data/blob/master/doc/api.md#version (mcData.version) instead ?
might be less confusing to the user

u9g:
> what do you think about returning the actually used version https://github.com/PrismarineJS/node-minecraft-data/blob/master/doc/api.md#version (mcData.version) instead ?
> might be less confusing to the user

done

rom1504 (approved):

PR #129, opened 2021-07-06
add tile entities dump

u9g:


rom1504:
as discussed in discord, let's remove tile entity data files

u9g:
@nickelpro

PR #131, opened 2021-08-07
Fix pe -> bedrock

nickelpro:
Tests are failing because trying to lookup PE versions in MCD now fails, this moves everything over to "bedrock"


nickelpro:
Bedrock 1.0 is failing tests but I don't think that's my fault lol

nickelpro:
Official ruling from @extremeheat is "Ya that's weird don't worry about it"

rom1504 (approved):

rom1504 (approved):

PR #132, opened 2021-08-07
ChunkSection is the same 1.14 -> 1.16, putting in common

rom1504:
noticed by @nickelpro


nickelpro (approved):
As discussed, LGTM. Correct `BitArray` used for each version

PR #133, opened 2021-08-08
1.17 Chunk Support Rewrite

nickelpro:
This is not good code. I would venture to say this is terrible code.

I Ctrl-C + Ctrl-V'd 1.16 and unga-bunga'd my way until the tests passed, this is the result.

In my defense, all of pr-chunk is janky as hell


rom1504:
congrats on making this work!
can you check if the API is the same in 1.16 and 1.17 ?
if not, can you check if it's possible to make it the same ? (even if some param should be passed as undefined in some version or similar ideas)
if really not possible, then the api doc https://github.com/PrismarineJS/prismarine-chunk#api should be adapted

src/pc/1.14/ChunkColumn.js
```diff
@@ -197,6 +199,14 @@ module.exports = (Block, mcData) => {
       return this.sectionMask
     }
 
+    getMaskArray () {
```
rom1504, line 202: is this new method useful ?

src/pc/1.14/ChunkColumn.js
```diff
@@ -197,6 +199,14 @@ module.exports = (Block, mcData) => {
...
+      return [this.getMask()]
+    }
+
+    countVerticalSections () {
```
rom1504, line 206: this should be added to api doc if it's useful

src/pc/1.14/ChunkColumn.js
nickelpro: No, it's a hold over from when I was rebasing from Archengius. Will remove

nickelpro: Not new, present in 1.16

nickelpro: I don't think it's useful either, but again I just Ctrl+C, Ctrl-V

src/pc/1.14/ChunkColumn.js
nickelpro: Again I would much rather just remove these methods, but they were present in 1.16 so for API consistency they're present here

src/pc/1.14/ChunkColumn.js
rom1504: https://github.com/PrismarineJS/prismarine-chunk/pull/133/files pretty sure it was added in this pr in all versions

src/pc/1.14/ChunkColumn.js
rom1504: same answer as other method, it was added in this pr

nickelpro:
API is consistent with 1.16 with the exception of:
* Bitmaps/masks are now long arrays
* Removed `minWorldHeight` and `maxWorldHeight` since they're not useful or meaningful in 1.17

nickelpro:
Now the only API inconsistency is the bitmap Number -> Array of protodef Longs, which I don't think is reducible complexity

Karang (approved):
LGTM

rom1504:
new exposed things and changes to API should be documented  in the readme

rom1504 (approved):

PR #144, opened 2021-11-10
[1.17] Fix ChunkColumn::getMask()

Saiv46:
`ChunkColumn::getMask()` method expected to return a number, not an array.


Saiv46:
And yeah, we can have up to 848 block-tall chunks before it'll break (1.18 limit is 384).

PR #148, opened 2021-12-27
Update 1.18 chunk dump

nickelpro:
Update the chunk dump to include lighting data from the main chunk packet


rom1504:
988 seems like too many files

rom1504:
988 seems like too many files

nickelpro:
It's adding 200 files and removing the 788 that are there currently, for a total of 988 files changed.

I'm _getting rid of_ 588 files

rom1504:
yeah indeed it was wrong before
it's still wrong after this pr though
the right number is 20 .dump files.
chunk dumper allows you to get exactly 20

nickelpro:
Ah, didn't realize we had a standard. I'll close and re-open with correct data

PR #150, opened 2022-01-01
1.18

nickelpro:
Completely borked WIP branch, presented without comment for the curious


nickelpro:
I expect tests to pass, but do not consider this mergable until the SingleValuePalette bug is better understood

nickelpro:
This is a mojang bug, I'm going to update the test comment and squash this commit. It should still get at least one code review before merge

Karang (approved):
LGTM

PR #162, opened 2022-01-13
Implement prismarine-registry, basic block entities and 1.18 disk loading

extremeheat:
depends on prismarine-block update

WIP, this goes along with updating anvil-provider for 1.18 level format


src/bedrock/0.14/chunk.js
```diff
@@ -17,13 +17,13 @@ const writeUInt4LE = require('uint4').writeUInt4LE
 
 module.exports = loader
 
-function loader (mcVersion) {
-  Block = require('prismarine-block')(mcVersion)
+function loader (registry) {
```
rom1504, line 20: same point as in prismarine-block, let's support both registry and version here

src/pc/1.15/ChunkColumn.js
```diff
@@ -190,10 +195,6 @@ module.exports = (Block, mcData) => {
       this.biomes[getBiomeIndex(pos)] = biome
     }
 
-    setBiomeColor (pos, r, g, b) {
```
rom1504, line 193: this needs to stay in existence due to the fact it exists in some mcpe (unless you remove these versions completely)

src/pc/common/PaletteContainer.js
```diff
@@ -127,7 +127,7 @@ class IndirectPaletteContainer {
 class SingleValueContainer {
   constructor (options) {
     this.value = options?.value ?? 0
-    this.bitsPerValue = options?.bitsPerValue ?? constants.MIN_BITS_PER_BLOCK
```
nickelpro, line 130: This is wrong, this value represents the `bitsPerValue` to use when _upgraded_ to a non-single value container. The default can be whatever, but you _must_ respect the option when passed.

extremeheat: Fixed

src/pc/1.17/ChunkColumn.js
```diff
@@ -153,11 +163,6 @@ module.exports = (Block, mcData) => {
       return this.biomes[getBiomeIndex(pos, this.minY)]
     }
 
-    getBiomeColor (pos) {
```
rom1504, line 156: can you undo this?

src/pc/1.15/ChunkColumn.js
nickelpro: They still exist, they've just been moved to `CommonChunkColumn` which this extends

src/pc/1.17/ChunkColumn.js
nickelpro: Did here, moved to `CommonChunkColumn`

src/pc/1.15/ChunkColumn.js
extremeheat: only the pc/* versions have the getBiomeColor/setBiomeColor moved to pc/common/CommonChunkColumn, so mcpe versions shouldn't be effected

rom1504:
bumped pblock

extremeheat:
Entity data should go into `.entity` property of a prismarine-block passed to getBlock/setBlock. Added some doc here for the `getBlockEntity(pos)` and `setBlockEntity(pos, nbt)` functions. If the block entities are sent along with the chunk data it should be possible to add in a extra parameter to the load() function, otherwise it can use loadBlockEntities(nbtArray)... 

Now since it looks like the "id" string field isn't written on the chunk packet that may have to be added in somewhere

rom1504:
ok sounds good for loadBlockEntities
what about dump ? I guess there should be a dumpBlockEntities ?

rom1504:
ok sounds good for loadBlockEntities
what about dump ? I guess there should be a dumpBlockEntities ?

rom1504:
ok sounds good for loadBlockEntities
what about dump ? I guess there should be a dumpBlockEntities ?

rom1504:
I'm thinking it would make sense to try to integrate this in mineflayer before merging
so we don't need to change the API after merging

nickelpro:
We could major version bump here, and then integrate to mineflayer and make any tweaks that come up. That way mineflayer maintains continuity of function.

extremeheat:
For mineflayer I think moving most of the relevant world data to prismarine-world would make sense to do now, to avoid things having to be reworked a second time

rom1504:
> We could major version bump here, and then integrate to mineflayer and
make any tweaks that come up. That way mineflayer maintains continuity of
function

My point was to avoid doing many major versions of prismarine-chunk because
we don't know what we need before integrating

rom1504:
> For mineflayer I think moving most of the relevant world data to
prismarine-world would make sense to do now, to avoid things having to be
reworked a second time

Yes let's do it.
I believe it will be cleaner to do it before we merge this. (Just do one PR
for mineflayer one for pworld and one for pchunk (this one))

nickelpro:
Sounds like a plan, would it be wise to move search at the same time? I can open a PR off this branch for it

rom1504:
missing chunk.dumpBlockEntities

PR #167, opened 2022-02-06
Block sections are not biomes

nickelpro:
Minor copy-paste bug fix

### DIFF
--- a/src/pc/common/PaletteChunkSection.js
+++ b/src/pc/common/PaletteChunkSection.js
@@ -79,9 +79,9 @@ class ChunkSection {
       data: palette.length === 1
         ? new SingleValueContainer({
             value: palette[0],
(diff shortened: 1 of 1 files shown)


extremeheat:
yeah I had a feeling the problem was around here...

rom1504:
How come it doesn't break any test :) ?
Can we easily add a non regression test ?

nickelpro:
I don't even know what this code is used for, I just spotted the error idly browsing diffs.

I'm not sure what to test for exactly, I guess you could write a stupid test that checks these exact variables are set to correct constants but that seems very narrow and fragile. It's not testing a broad function, just that these constants were set correctly

rom1504:
I mean, what is the consequence of having incorrect values for bitsPerValue and maxBits for singleValueContainer?
shouldn't it break something?

rom1504:
Ok

PR #168, opened 2022-02-06
Bedrock 1.16 - 1.18.0 chunks

extremeheat:
Implements Bedrock Edition chunk format support for versions 1.16.220 - 1.18.0, with cache/no cache handling


test/ChunkColumn.test.js
```diff
@@ -1,106 +1,144 @@
 /* eslint-env mocha */
 
-const Block = require('prismarine-block')('1.13')
-const mcData = require('minecraft-data')('1.13.2')
-const ChunkColumn = require('../src/pc/1.13/ChunkColumn')(Block, mcData)
+// const versions = ['bedrock_0.14', 'bedrock_1.0', 'bedrock_1.18.0', '1.8', '1.9', '1.10', '1.11', '1.12', '1.13.2', '1.14.4', '1.15.2', '1.16.1', '1.17', '1.18']
```
rom1504, line 3: we should probably test them all here indeed

rom1504:
cool!

if you have some chunk dumps, it could be useful to put them in the test folder, so the cycle test can run here

rom1504:
passing tests, but I still think it would be better with some chunk dumps so cycle tests could run

extremeheat:
working on it, for now using bedrock-provider's existing tests (it handles the chunk network protocol) to do the dumps, then can be imported here

extremeheat:
Added some network tests here, in a seperate test file since bedrock has several ways to receive chunks

The test files for pchunk are dumped after running the tests in bedrock-provider (which also tests the chunk network decoding process) - https://github.com/extremeheat/bedrock-provider/blob/9b5d2b5568f129b3cceb00120b2925f67692b7d1/test/worldLoad.ts#L27 ... since they're both linked both need to be updated about the same time whenever there's a major update

Still have some tests to do here and in bedrock-provider related to heightmaps and entities (although I think entities can be handled next time) 

Current tests here are failing because of caching and [xxhash-wasm requiring nodejs 16](https://github.com/jungomi/xxhash-wasm/pull/26), the old 0.x version of it does support node 14 but it had performance issues and didn't handle little/big endian correctly so I think the options here are to either:
* introduce some global polyfill to make it work in node 14 (idk if this will work yet)
* use 0.x version of xxhash-wasm and correct endian issues through some bitwise operations

Also related to caching, need to do some thinking about the best way to handle it.
* some of the caching implementation files here (BlobType, BlobEntry, etc) may make sense in a different package since they're not related to minecraft but are general cache handling functions. 
* maybe would make sense for cache storage/retrieval to be put into prismarine-registry

rom1504:
I notice some of the dumped files have undefined in their name. Is that correct?

extremeheat:
yeah, that's just because of version differences in 1.18 and below (1.18 has a y field, but if it doesn't exist it gets stringified to undefined)

rom1504:
I'm ok with moving everything to node 16, it's lts now

extremeheat:
actually hmm no, I think that's not intentional for level_chunk

extremeheat:
failing now since the ChunkColumn.js tests are run on all versions, some versions are missing state IDs. I think a simple fix for that would just be to assign a stateId for all versions in prismarine-block or node-minecraft-data, on versions <1.12 by combining the type ID and metadata fields (`blockType << 4 | blockMetadata`)

extremeheat:
Fixed the pc issues here by fixing 1.8 initialization function not retuning correctly and 1.18 not loading block entities in constructor, and with https://github.com/PrismarineJS/prismarine-block/pull/55 

For the bedrock side i haven't tested this in game and with prismarine viewer yet but once that's done via some extra tests in bedrock provider I think this should be ready

rom1504:
since this is pretty big, would appreciate additional reviews here from @u9g @nickelpro @Karang

src/bedrock/common/PalettedStorage.js
```diff
@@ -0,0 +1,87 @@
+const wordByteSize = 4
+const wordBitSize = wordByteSize * 8
+
+class PalettedStorage {
+  constructor (bitsPerBlock) {
+    this.bitsPerBlock = bitsPerBlock
+    this.blocksPerWord = Math.floor(wordBitSize / bitsPerBlock)
+    // this.paddingPerWord = wordBitSize % bitsPerBlock
```
u9g, line 8: did you mean to leave this?

package.json
```diff
@@ -38,16 +38,19 @@
...
+    "prismarine-block": "^1.14.1",
     "prismarine-registry": "^1.1.0",
     "smart-buffer": "^4.1.0",
     "uint4": "^0.1.2",
-    "vec3": "^0.1.3"
+    "vec3": "^0.1.3",
+    "xxhash-wasm": "^1.0.1"
   },
   "engines": {
     "node": ">=14"
```
u9g, line 56: if the ci requires node 16, should this too?

u9g (changes requested):
Small fixes, otherwise looks good assuming it works when testing with prismarine-viewer :)

extremeheat:
OK, I tested with minecraft client on 1.17 and it everything looks fine, going to assume it works on 1.18 and 1.16 based on tests. I haven't tested prismarine viewer as it's missing some updated assets but I assume that should work too. So this is ready for review, if any problems arise I think they can be addressed in a separate PR to not make this any bigger

extremeheat:
ok.. so with respect to node 16, it seems colab is stuck at node 14 for now... and the hasher doesn't work on node 14 because a wasm error 
image

... apparently wasm i64s are only supported in newer v8 versions and not on the v8 version node 14 is using, which is news to me

https://v8.dev/features/wasm-bigint

So using v0.x and building the BigInt in a hacky way like done in protodef may be necessary

README.md
```diff
@@ -32,15 +33,23 @@ console.log(JSON.stringify(chunk.getBlock(new Vec3(3,50,3)),null,2))
...
 Test data can be generated with [minecraftChunkDumper](https://github.com/PrismarineJS/minecraft-chunk-dumper).
 
 Install it globally with `npm install minecraft-chunk-dumper -g` then run :
 
 `minecraftChunkDumper saveChunk 1.8.8 1.8/chunk.dump 1.8/chunk.meta`
 
-## API
+### bedrock
+
+Run tests in [bedrock-provider](https://github.com/PrismarineJS/bedrock-provider/) and copy data from `fixtures/$version/pchunk` into `tests/bedrock_$version`
```
rom1504, line 46: CCan

extremeheat: The tests in bedrock-provider are done in combination with a client and server, since the server only loads and saves chunks after a player joins the server. The client tests there decode the chunks with the ChunkColumn implementation that was moved to pchunk, but doesn't do any special tests beyond that as to if network format re-encoding works or other chunk API methods.

README.md
```diff
@@ -211,3 +231,14 @@ is in this section.
 #### setSkyLight(pos: Vec3, light: number): void
 #### dump(): Buffer
 #### load(data: Buffer, skyLightSent?: boolean): void
+
+### bedrock
+
+See [index.d.ts](https://github.com/PrismarineJS/prismarine-chunk/blob/master/types/index.d.ts#L56)
+
+#### compact()
```
rom1504, line 239: I'm not sure if we should really expose that but ok for now

src/bedrock/common/Stream.js
```diff
@@ -0,0 +1,364 @@
+const MAX_ALLOC_SIZE = 1024 * 1024 * 2 // 2MB
+const DEFAULT_ALLOC_SIZE = 10000
+
+class ByteStream {
```
rom1504, line 4: this looks a lot like protodef..

rom1504:
ok lgtm, let's go

src/bedrock/common/Stream.js
extremeheat: Yeah, but I needed zigzag varints, peek and buffer concatenation functions. The original lib used in bedrock-provider was the JSPrismarine/BinaryStream class, but that did a bunch of unnecessary slow assertions on top of the ones in the node Buffer class

PR #170, opened 2022-02-19
INCOMPLETE: Typescript typings.

GenerelSchwerz:
I have written some rough types for all of prismarine-chunk using what was there prior. The previous typings did not expose the Chunk class as a type, so return values for prismarine-world would always return as any, if properly typed.

The shortcoming of these typings is that the typings are not version-specific. Rather, they represent the CommonChunkColumn class.


rom1504:
just merged a big pr
can you rebase/check if this is still relevant?

TRCYX:
Seems unnecessary. `InstanceType<C>` can retrieve a class out of the type of its constructor. Also, the PR mentioned above changed a lot of typing, so I believe it's fine to close this.

rom1504:
old, please reopen if you want to finish

Issue #183, opened 2022-05-28
PaletteStorage.incrementPalette can go out of bounds


lowlines:
https://github.com/PrismarineJS/prismarine-chunk/blob/0f4ec64bd0ef8a84f04d825358dd1443e4310aa6/src/bedrock/common/PalettedStorage.js#L95-L102

If you have bitsPerBlock size that causes rounding on the wordCount, this double for loop will go beyond the 4096 block range and throw and error.

extremeheat:
Please open PR if you have a fix.

PR #184, opened 2022-06-04
PaletteStorage.incrementPalette fix

lowlines:
See https://github.com/PrismarineJS/prismarine-chunk/issues/183


nickelpro:
This looks like an appropriate fix to me, pinging @extremeheat

Also rom is going to want a test case that fails on the current code so that this doesn't get reintroduced from a future re-factor or something like that.

extremeheat:
OK, will defer to nickelpro, LGTM

Is there a reason to be exposing the internal classes top level? This does not sound right, it's not a documented part of the package

lowlines:
Exposing the internal classes is unrelated to this fix. Feel free to ignore those changes.

extremeheat:
Can you fix? (If not @nickelpro should have write perms to fix)

extremeheat (approved):

nickelpro:
I only verified the actual fix. My eyes glazed over the exposure/TS stuff 😅

I assume people know what they're doing with typings because I certainly don't

extremeheat:
I'm was thinking we could prevent this by changing for loop condition, but can't think of a fix that way.

LGTM for merge @nickelpro

nickelpro:
LGTM @rom1504

nickelpro:
This still needs a test case ya!

rom1504:
please add a test

rom1504:
conflict, please resolve and re open if you want

PR #204, opened 2022-11-08
add helper function for loading blockentities directly from packet data

qrvd:
The `blockEntities` array in the [map_chunk](https://minecraft-data.prismarine.js.org/?d=protocol#toClient_map_chunk) packet contains NBT objects that can't be used with `ChunkColumn.loadBlockEntities`. They look like this:

```js
// example of what "blockEntities" can look like
[
  {
    type: 'compound',
    name: '',
    value: { x: [Object], y: [Object], z: [Object], id: [Object] }
  }
]
```
These can't be directly passed to `ChunkColumn.loadBlockEntities`, because the x, y, z coordinates are not stored at the top level of the object. On the other hand, the new method `loadNBTBlockEntities` can be used with this value just fine.

### DIFF
--- a/src/pc/common/CommonChunkColumn.js
+++ b/src/pc/common/CommonChunkColumn.js
@@ -61,9 +61,19 @@ class CommonChunkColumn {
     delete this.blockEntities[posKey(pos)]
   }
 
+  loadBlockEntity (entity) {
+  this.setBlockEntity({ x: entity.x.value >> 4, y: entity.y.value, z: entity.z.value >> 4 }, entity)
+  }
+
+  loadNBTBlockEntities (entities) {
+    for (const nbt of entities) {
+      this.loadBlockEntity(nbt.value);
+    }
+  }
+
   loadBlockEntities (entities) {
     for (const entity of entities) {
-      this.setBlockEntity({ x: entity.x.value >> 4, y: entity.y.value, z: entity.z.value >> 4 }, entity)
+      this.loadBlockEntity(entity);
     }
   }
 }


extremeheat:
How is this handled in mineflayer currently ?

qrvd:
@extremeheat Mineflayer manually adds each block entity.
https://github.com/PrismarineJS/mineflayer/blob/803024cbc348276abc780c0d7ec810b3740e5782/lib/plugins/blocks.js#L290
Apparently the block entity objects have changed since 1.17 and there are no problems with newer versions, but the helper function would benefit users of earlier versions of Minecraft. I'm using 1.16.5 so that's why I ran into this problem.

extremeheat:
And what's wrong with that? Per that link the protocol structure is not the same on all versions.

qrvd:
> And what's wrong with that? Per that link the protocol structure is not the same on all versions.

Well the code works. But the loadBlockEntities method goes unused since it can't be used either for the older or the newer version of the packet.

extremeheat:
IIRC, loadBlockEntities is for disk loading in avil level provider, not network

extremeheat:
Yes, the load function is just used by anvil provider. https://cs.github.com/PrismarineJS/prismarine-provider-anvil/blob/e5814c0fc139fdc6b419f846d8a7d4083fe07edf/src/1.18/chunk.js#L134

As shown in your mineflayer link, the exact code for loading block entities array varies by version and is pretty trivial so I think this makes sense to handle in mineflayer over extra functions in prismarine-chunk.

rom1504:
old and failing

would be great though

PR #208, opened 2023-02-05
Fix loading skylight when in nether

### LINKED ISSUE #186: The value of "offset" is out of range. It must be >= 0 and <= 10239. Received 10684
The bot is crashing when joining the server. I dont have the exact code this is happening with.
However, I do know that this is being caused by a sign in the nether, which when destroyed it the bot wont crash while logging in, when replaced it will crash

stacktrace
https://sourceb.in/1iB0fA68TD

frej4189:
This fixes a long-standing mineflayer issue (https://github.com/PrismarineJS/mineflayer/issues/2614) that would cause the bot to crash when coming across a sign in the nether

Closes #186

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ Get the block light at `pos`
 
 #### Chunk.getSkyLight(pos)
 
-Get the block sky light at `pos`
+Get the block sky light at `pos`. Undefined when world does not have skylight (nether or end in Vanilla)
 
 #### Chunk.getBiome(pos)
 
@@ -223,7 +223,7 @@ See [index.d.ts](https://github.com/PrismarineJS/prismarine-chunk/blob/master/ty
 #### getBlockType(pos: Vec3): number
 #### getBlockData(pos: Vec3): number
 #### getBlockLight(pos: Vec3): number
-#### getSkyLight(pos: Vec3): number
+#### getSkyLight(pos: Vec3): number | undefined
 #### setBlockStateId(pos: Vec3, stateId: number): void
 #### setBlockType(pos: Vec3, id: number): void
 #### setBlockData(pos: Vec3, data: Buffer): void
--- a/src/pc/1.13/ChunkSection.js
+++ b/src/pc/1.13/ChunkSection.js
@@ -150,7 +150,7 @@ class ChunkSection {
   }
 
   getSkyLight (pos) {
-    return this.skyLight ? this.skyLight.get(getBlockIndex(pos)) : 0
+    return this.skyLight ? this.skyLight.get(getBlockIndex(pos)) : undefined
   }
 
   setBlockLight (pos, light) {
--- a/src/pc/1.8/chunk.js
+++ b/src/pc/1.8/chunk.js
@@ -110,7 +110,7 @@ module.exports = (registry) => {
     }
 
     getSkyLight (pos) {
-      if (!this.skyLightSent) return 0
+      if (!this.skyLightSent) return undefined
       const section = this._getSection(pos)
       return section ? section.getSkyLight(posInSection(pos)) : 15
     }
--- a/src/pc/1.9/ChunkSection.js
+++ b/src/pc/1.9/ChunkSection.js
@@ -151,7 +151,7 @@ class ChunkSection {
   }
 
   getSkyLight (pos) {
-    return this.skyLight ? this.skyLight.get(getBlockIndex(pos)) : 0
+    return this.skyLight ? this.skyLight.get(getBlockIndex(pos)) : undefined
   }
 
   setBlockLight (pos, light) {


rom1504:
This should be updated in the doc

Why is undefined better than 0 ?

frej4189:
> This should be updated in the doc
> 
> Why is undefined better than 0 ?

Because 0 is an actual skylight value. We need it to be undefined so we don't check for skylight in nether

nickelpro:
Does this fix the crash? The problem in Mineflayer is calls to `setSkyLight()` not `getSkyLight()`?

The problem seems to be if you call `setSkyLight()` we blindly believe you that a skylight exists, and instead the Chunk need to check `.skyLightSent` before calling into the section.

frej4189:
> Does this fix the crash? The problem in Mineflayer is calls to `setSkyLight()` not `getSkyLight()`?
> 
> The problem seems to be if you call `setSkyLight()` we blindly believe you that a skylight exists, and instead the Chunk need to check `.skyLightSent` before calling into the section.

The problem is that the blocks returned from getBlock will include skylight data even when there is none. That causes an issue when the same block is then used in a setBlock call later.

nickelpro:
Ie, this is wrong: https://github.com/PrismarineJS/prismarine-chunk/blob/90704e1ec07fe89e048fdc07221f78d9ccec2b47/src/pc/1.8/chunk.js#L143-L146

We can't blindly call into sections like this, because the structure of a section is dependent on the presence/absence of a skylight

nickelpro:
> > Does this fix the crash? The problem in Mineflayer is calls to `setSkyLight()` not `getSkyLight()`?
> > The problem seems to be if you call `setSkyLight()` we blindly believe you that a skylight exists, and instead the Chunk need to check `.skyLightSent` before calling into the section.
> 
> The problem is that the blocks returned from getBlock will include skylight data even when there is none. That causes an issue when the same block is then used in a setBlock call later.

Sure, but this isn't wrong. A skylight of zero is expected on a block that comes from a dimension that does not have a skylight. I think the right answer is that subsequent calls to `setSkyLight()` on such a block are discarded.

frej4189:
> > > Does this fix the crash? The problem in Mineflayer is calls to `setSkyLight()` not `getSkyLight()`?
> > > The problem seems to be if you call `setSkyLight()` we blindly believe you that a skylight exists, and instead the Chunk need to check `.skyLightSent` before calling into the section.
> > 
> > 
> > The problem is that the blocks returned from getBlock will include skylight data even when there is none. That causes an issue when the same block is then used in a setBlock call later.
> 
> Sure, but this isn't wrong. A skylight of zero is expected on a block that comes from a dimension that does not have a skylight. I think the right answer is that subsequent calls to `setSkyLight()` on such a block are discarded.

Sure. The issue can also be fixed by replacing the call to exists in setBlock with a check that skylight is non-zero, but then we are won't be updating sky-light in the overworld.

nickelpro (approved):
Discussed in Discord, I agree with the change to `undefined`

This should be merged
@rom1504

rom1504:
Does this apply to other versions? What about other properties?

frej4189:
> Does this apply to other versions? What about other properties?

I have updated it for all versions. Skylight is the only property that does not exist universally in chunks

nickelpro:
Looks like this would still be an issue in 1.14 unless I'm missing something, and we introduce a mild inconsistency with 1.15+ which still `return 0` here, but they don't really have a concept of a "nonexistent" skylight

frej4189:
> Looks like this would still be an issue in 1.14 unless I'm missing something, and we introduce a mild inconsistency with 1.15+ which still `return 0` here, but they don't really have a concept of a "nonexistent" skylight

1.14+ uses separate sections for blocks and skylight, so the check for existing skylight sections that is currently in place fixes the issue for those versions

extremeheat:
Returning undefined would be a breaking change. If some dimensions do not have lighting then it's something that should be handled inside the user code, not individually on every getSkyLight() call.

frej4189:
> Returning undefined would be a breaking change. If some dimensions do not have lighting then it's something that should be handled inside the user code, not individually on every getSkyLight() call.

It is breaking to the API, but calling getSkyLight on a chunk that does not have skylight should not be allowed. See Discussion in Discord

extremeheat:
Re - https://discord.com/channels/413438066984747026/743199746876768318/1071819420919414795

I agree with @nickelpro that it's something for user code to handle. If the sky light isn't sent, it should be handled in the calling code in an outer block, not inside a for loop iteration which is where more often this code will be called (for example with rendering in prismarine-viewer). Blocks returning undefined is a expected scenario if a block doesn't exist at a specific position. As in Minecraft, all bounded positions in a chunk have light values whether they are occupied or not. We don't currently compute light in a chunk but whether or not we did, just like in Minecraft, the world will appear either fully dark until light is loaded from the server or computed. For the context of this PR, IMO it doesn't make sense to handle this per-call in a breaking manner like this (it's not done on bedrock either).

nickelpro:
The question is really, _who is this breaking for_? Pris-chunk somewhat expects these values to be undefined when you hand it a `Block` via `setBlock()`. The nature of the bug is that you cannot round-trip `getBlock()` -> `setBlock()` in 1.8 thru 1.13.

I agree with @frej4189 insomuch as whatever we do with `getBlock()` should be reflected in what we accept at `setBlock()`. To make that happen, any change will be "breaking", but I'm unconvinced that anyone consumes this API in such a manner except for the `update_sign()` routine of MF's blocks.js.

frej4189:
Redoing

PR #209, opened 2023-02-05
Fix setBlock calling setSkylight when chunk does not have skylight

### LINKED ISSUE #186: The value of "offset" is out of range. It must be >= 0 and <= 10239. Received 10684

frej4189:
Closes #186

### DIFF
--- a/src/pc/1.13/ChunkColumn.js
+++ b/src/pc/1.13/ChunkColumn.js
@@ -74,7 +74,7 @@ module.exports = (Block, mcData) => {
       if (typeof block.biome !== 'undefined') {
         this.setBiome(pos, block.biome.id)
       }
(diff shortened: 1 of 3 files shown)


frej4189:
Appears to break 1.19+ for some odd reason

frej4189:
> Appears to break 1.19+ for some odd reason

Might just be nether being broken in 1.19+ in general (including in current release).

extremeheat:
Possibly, that will need to be fixed

extremeheat:
There should be a test here also

PR #226, opened 2023-06-25
1.20

PondWader:
After https://github.com/PrismarineJS/minecraft-data/pull/723 for tests because of the state ID offset issue causing issues with reading some blocks


test/test.js
```diff
@@ -21,19 +21,20 @@ versions.forEach((version) => describe(`Chunk implementation for minecraft ${ver
 
   const isPostFlattening = version.startsWith('1.13') || version.startsWith('1.14') ||
```
extremeheat, line 22: These test cases should probably be using mc-data features

rom1504: Indeed

Issue #164, opened 2022-01-25
Error reading direct palettes for 1.18


u9g:
```
Error: Attempted to read beyond the bounds of the managed data.
    at SmartBuffer.ensureReadable (C:\Users\---\Documents\js-projects\my-bot\node_modules\smart-buffer\build\smartbuffer.js:1093:19)
    at SmartBuffer._readNumberValue (C:\Users\---\Documents\js-projects\my-bot\node_modules\smart-buffer\build\smartbuffer.js:1163:14)
    at SmartBuffer.readUInt32BE (C:\Users\---\Documents\js-projects\my-bot\node_modules\smart-buffer\build\smartbuffer.js:361:21)
    at BitArray.readBuffer (C:\Users\---\Documents\js-projects\my-bot\node_modules\prismarine-chunk\src\pc\common\BitArrayNoSpan.js:183:34)
    at DirectPaletteContainer.readBuffer (C:\Users\---\Documents\js-projects\my-bot\node_modules\prismarine-chunk\src\pc\common\PaletteContainer.js:30:15)
    at Function.read (C:\Users\---\Documents\js-projects\my-bot\node_modules\prismarine-chunk\src\pc\common\PaletteChunkSection.js:96:12)
    at ChunkColumn.load (C:\Users\---\Documents\js-projects\my-bot\node_modules\prismarine-chunk\src\pc\1.18\ChunkColumn.js:242:41)
    at addColumn (C:\Users\---\Documents\js-projects\my-bot\node_modules\mineflayer\lib\plugins\blocks.js:99:14)
    at Client.<anonymous> (C:\Users\---\Documents\js-projects\my-bot\node_modules\mineflayer\lib\plugins\blocks.js:319:5)
    at Client.emit (node:events:394:28)
```

rom1504:
when

MetuMortis-code:
on joining server

u9g:
still getting ```
RangeError: Invalid typed array length: Infinity
    at new Uint32Array (<anonymous>)
    at new BitArray (C:\Users\Jason\Documents\js-projects\docs-bot\node_modules\prismarine-chunk\src\pc\common\BitArrayNoSpan.js:16:17)
    at new DirectPaletteContainer (C:\Users\Jason\Documents\js-projects\docs-bot\node_modules\prismarine-chunk\src\pc\common\PaletteContainer.js:8:17)
[...]

PR #232, opened 2023-08-24
Fix direct palettes

frej4189:
Closes https://github.com/PrismarineJS/prismarine-chunk/issues/164

There are a handful of problems with the implementation of direct palettes that would cause errors on some servers (most notably Hypixel).

This PR addresses these issues:

1. The BitArray readBuffer function always attempts to consume data enough to populate the entire storage. This will cause an error when the server provides a shorter (or longer) long array than needed.
2. The direct palette uses the server provided bits per value, instead of truncating to the size of the registry.

I generally believe these changes to be correct, but I have ***not*** tested any versions below 1.18. I'm putting up the PR now so others can test with mineflayer and so we can get reviews started.


extremeheat:
> Also CI won't pass most tests since it relies on dumps generated with the incorrect implementation.

What do you mean by "incorrect implementation"? The test data should just be desterilized chunk data off the network, no?

frej4189:
> > Also CI won't pass most tests since it relies on dumps generated with the incorrect implementation.
> 
> What do you mean by "incorrect implementation"? The test data should just be desterilized chunk data off the network, no?

Dumps seem to be generated by pchunk itself

rom1504:
They are not. Dumps are raw bytes generated using Minecraft-protocol

The tests using them check internal consistency of pchunk between load and dump

So if they're not passing , something is broken

frej4189:
I see.

I haven't fully implemented dumping so that could be the cause of this. I just saw invalid data in the dumps (namely direct palette with bit per block set to 16) but that may well be an issue with the dumping method if it compares the new dumps with the ones from the network.

rom1504:
If data in the dumps is invalid it means there is a bug in the vanilla Minecraft server implementation

frej4189:
Yeah it's all good. I'm just saying there's something wrong with my dump method. Not that there's something wrong with the dumps

test/test.js
```diff
@@ -14,7 +14,7 @@ const expect = require('expect').default
 const versions = ['bedrock_0.14', 'bedrock_1.0', '1.8', '1.9', '1.10', '1.11', '1.12', '1.13.2', '1.14.4', '1.15.2', '1.16.1', '1.17', '1.18', '1.19', '1.20']
 const cycleTests = ['1.8', '1.9', '1.10', '1.11', '1.12', '1.13.2', '1.14.4', '1.15.2', '1.16.1', '1.17', '1.18', '1.19', '1.20']
 
-versions.forEach((version) => describe(`Chunk implementation for minecraft ${version}`, () => {
+versions.filter(version => version === '1.20').forEach((version) => describe(`Chunk implementation for minecraft ${version}`, () => {
```
lkwilson, line 17: Is this supposed to be here?

frej4189: Absolutely not. Forgot to remove that from my own tests.

src/pc/1.18/ChunkColumn.js
```diff
@@ -14,9 +15,11 @@ module.exports = (Block, mcData) => {
       this.minY = options?.minY ?? 0
       this.worldHeight = options?.worldHeight ?? constants.CHUNK_HEIGHT
       this.numSections = this.worldHeight >> 4
+      this.maxBitsPerBlock = neededBits(Object.values(mcData.blocks).reduce((high, block) => Math.max(high, block.maxStateId), 0))
+      this.maxBitsPerBiome = neededBits(Object.values(mcData.biomes).length)
```
lkwilson, line 19: Was this line left over from testing?

frej4189: No. This is needed for proper direct palette parsing

lkwilson: One of those lines is not needed though, right? It seems the rhs has no side effects and the second statement overwrites the computed value from the first.

lkwilson: Ah, perBlock and perBiome, nvm I see it now

frej4189: One is bits per block. The other is bits per biome

extremeheat:
Cc @nickelpro if you have the time to review

src/pc/common/PaletteBiome.js
nickelpro, line 1: Is this safe to remove? Aren't we expected to exhaust this buffer?

src/pc/common/PaletteBiome.js
```diff
@@ -83,7 +82,6 @@ class BiomeSection {
       palette.push(varInt.read(smartBuffer))
     }
 
-    varInt.read(smartBuffer)
```
nickelpro, line 86: Is this safe to remove, are we expected to exhaust this buffer?

frej4189: I just moved it into the PaletteContainer

nickelpro:
Other than the one comment, which might be nonsense because I'm not 100% clear on why that `read` was there to begin with, LGTM **WRT** the commit doing what it says it does.

I think there's a bit of repetition here that could use refactoring, but the basic idea of only using `GLOBAL_BITS_PER_X` when we know that it's big enough is correct and the current behavior should be considered a bug.

I'm less certain about trusting servers about their long counts, but testing should reveal the safety of that.

src/pc/common/BitArray.js
```diff
@@ -134,6 +134,7 @@ class BitArray {
   readBuffer (smartBuffer, size = this.data.length) {
     if (size !== this.data.length) {
       this.data = new Uint32Array(size)
+      this.readOffset = size * 4
```
nickelpro, line 137: This is crazy surprising, I don't think this belongs here.

If I call `readBuffer` a read should always happen. I might be missing context

frej4189:
I don't think there's anymore repetition than earlier so I don't think that's too much of an issue.

With regards to trusting servers long counts, that is also what the Vanilla client does. See https://github.com/extremeheat/extracted_minecraft_data/blob/20e005fafc4928179f144d3ff68413f0ca70808b/client/net/minecraft/network/FriendlyByteBuf.java#L402-L417

src/pc/common/BitArray.js
frej4189: Check my comment here https://github.com/PrismarineJS/prismarine-chunk/pull/232#issuecomment-1693890460

frej4189: And for additional context see https://github.com/extremeheat/extracted_minecraft_data/blob/20e005fafc4928179f144d3ff68413f0ca70808b/client/net/minecraft/world/level/chunk/PalettedContainer.java#L163

nickelpro: I don't see how return'ing early from `readBuffer` after skipping the data that was requested to be read has anything to do with the linked code.

frej4189: To summarize what is happening in the Vanilla implementation:

The Vanilla client attempts to read a long array into the existing raw array in the ChunkStorage. The readLongArray function then checks to see if the size of the existing array is equal to the size of the network provided long array (which is what I am doing with `size !== this.data.length` here). If it is not equal, a new long array will be created locally and returned from the function, but the return value of readLongArray is not used in the PalettedContainer read method, so the result is simply discarded.

My implementation technically doesn't mirror this behaviour as it doesn't create a new long array, but the result is the same, just less wasteful.

nickelpro:
> I don't think there's anymore repetition than earlier so I don't think that's too much of an issue.

prisChunk being bad is the norm, we do try to improve it from time-to-time. Agree not a blocker.

> With regards to trusting servers long counts, that is also what the Vanilla client does.

I'm down for this in principle, I'm just skeptical of modded servers in general to follow vanilla behavior and this entire patch series is about modded server support. Vanilla works fine.

src/pc/common/BitArray.js
nickelpro: Thank you for explaining

frej4189:
> I'm down for this in principle, I'm just skeptical of modded servers in general to follow vanilla behavior and this entire patch series is about modded server support. Vanilla works fine.

While I agree this technically is only a problem on modded servers, the current chunk implementation does in fact not mirror that of the Vanilla client. I am not jumping through hoops to support non-Vanilla servers, I am simply mirroring the Vanilla (client) implementation.

Also to add the current implementation actually does not work on Vanilla servers due to the GLOBAL_BITS_PER_BLOCK constant being incorrect for some versions. It just doesn't cause errors because we seldom fall back on that value.

nickelpro (approved):
Comments all resolved, I'm happy with this.

Holding off on merge for further field testing/rom's thumbs up

rom1504:
looks ok but yeah we really need to put more code in common between versions here or it's going to become very hard to maintain fast

rom1504:
maybe let's wait a few more days if anyone else wants to take a look (maybe @Karang ?)

frej4189:
Will add chunk dumps from Hypixel to test data aswell. I have a few that I tested locally but I haven't extracted it just the way the other dumps are generated so I will have to make some changes.

frej4189:
Added hypixel dumps and made a few changes to the dump tests to account for non-Vanilla format (i.e. histogram will not look like a naturally generated world). I don't think anymore testing is needed, but if anyone has comments to the code itself, feel free.

test/test.js
```diff
@@ -376,7 +392,7 @@ versions.forEach((version) => describe(`Chunk implementation for minecraft ${ver
           }
         }
 
-        if (!version.startsWith('1.8')) {
+        if (!version.startsWith('1.8') && !chunkDump.includes('hypixel')) {
```
rom1504, line 391: which are hypixel chunks not the same after dumping?

frej4189: Because hypixel provides excess data which is dropped by pchunk (per this PR). See my comment here https://github.com/PrismarineJS/prismarine-chunk/pull/232#discussion_r1306124496

Since the test already iterates over all blocks and palettes in the chunk, it means the two chunks are still functionally the same if the test passes

rom1504:
@frej4189 if you run the tests with implementation from `main` on hypixel chunks, does it fail? (eg does this PR fixes it)

frej4189:
> @frej4189 if you run the tests with implementation from `main` on hypixel chunks, does it fail? (eg does this PR fixes it)

It does indeed.

frej4189:
Running current implementation CI here https://github.com/frej4189/prismarine-chunk/pull/1

rom1504:
ok sounds good, thanks for the pr!

PR #237, opened 2024-03-24
Support for hashed block network ID's

FreezeEngine:


src/bedrock/1.18/SubChunk.js
```diff
@@ -3,6 +3,16 @@ const { StorageType } = require('../common/constants')
 const PalettedStorage = require('../common/PalettedStorage')
 
 class SubChunk118 extends SubChunk13 {
+  loadRuntimePalette (storageLayer, stream, paletteSize) {
+    this.palette[storageLayer] = []
+
+    for (let i = 0; i < paletteSize; i++) {
+      const runtimeId = stream.readZigZagVarInt()
+      const block = this.registry.blocksByRuntimeId[runtimeId]
+      this.palette[storageLayer][i] = { stateId: runtimeId, ...block, count: 0 }
```
extremeheat, line 12: `stateId` and `runtimeId` are not the same, stateId should be set to `block.stateId`, `runtimeId` should be the runtimeId (that should also be inside the block object return by this.registry.blocksByRuntimeId[runtimeId])

FreezeEngine: Should block.runtimeId be added to pblock? Or I can just do stateId: block.stateId? this.registry.blocksByRuntimeId = data.blockStates when hashes are not used

extremeheat: blocksByRuntimeId should look the exact same as blocksByStateId with the only difference being the key for the map being a runtimeID as opposed to a stateID

FreezeEngine: block.stateId doesnt seem to exsit, I can add it to pregestry by blocksByRuntimeId[pBlock.getHash(name, states)].stateId = stateId but then data.blocksByRuntimeId = data.blocksByStateId would not have that field, how should this be resolved?

extremeheat: Hmm, it seems node-minecraft-data doesn't expose a `.stateId`. So it seem the options to fix are:
1. add `.stateId` field in node-mcdata on top of `.minStateId` / `.defaultState` (this kind of feels wrong as typically node-mcdata maps right from blocks.json)
2. add in prismarine-registry a `.stateId` field to blocks objects inside blocksBy* (would require `.map` copying over mc-data block data)
3. add some separate map inside p-registry to map stateID <-> runtimeID (kind of messy to have many of these maps)

I'm not sure exactly best approach but 2 seems least problematic, cc @rom1504

FreezeEngine: I can do second for now, then could be reviewed and decided

rom1504:
what is the state of this? do you know @extremeheat ?

FreezeEngine:
Should be working

extremeheat:
Blocked by https://github.com/PrismarineJS/prismarine-registry/pull/33

rom1504:
I am ok with this, but would be nice to have tests

as is we don't know if it works as far as pchunk is concerned

PR #238, opened 2024-04-17
fix: toJson / fromJson didn't use maxBitsPerBlock

zardoy:
Fixes the following code:

```ts
import PChunk from 'prismarine-chunk'
import { Vec3 } from 'vec3'

const Chunk = PChunk('1.18.1')
const chunk = new Chunk(undefined)

const chunk2 = Chunk.fromJson(chunk.toJson()) as any

for (let i = 0; i < 4096; i++) {
    chunk2.setBlockStateId(new Vec3(0, 0, 0), i) // Decides to switch to Direct pallete at some point
    const blockStateId = chunk2.getBlockStateId(new Vec3(0, 0, 0))
    if (blockStateId !== i) throw new Error(`Expected ${i} but got ${blockStateId}`) // Expected 256 but got 0
}
```

Also what about indirect pallete I don't really know what is the correct behavior but it just feels like always adding a new block into pallete on the block replace is not right (since the block doesn't exist in chunk anymore).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist/
 benchmarks/results/
 package-lock.json
 .vscode
+.DS_Store
--- a/src/pc/common/PaletteContainer.js
+++ b/src/pc/common/PaletteContainer.js
@@ -113,6 +113,7 @@ class IndirectPaletteContainer {
       type: 'indirect',
       palette: this.palette,
       maxBits: this.maxBits,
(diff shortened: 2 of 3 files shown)


extremeheat:
Can you add a test?

zardoy:
> Can you add a test?

Of course, just wanted initial feedback. Also, is it a bug that Pallete doesn't get cleared? I feel its not optimized...

PR #239, opened 2024-06-07
fix loadParsedLights type!

zardoy:
will fix implementation later


rom1504:
In the future, always provide information on what you are fixing.

Here it seems, it's about adding that "?"

zardoy:
> In the future, always provide information on what you are fixing.
> 
> Here it seems, it's about adding that "?"

Ok, but I think it was obvious from the diff: the type definitions had the method that doesn't exist. And yeah, it is optional since it is present only in two last versions.

PR #240, opened 2024-06-07
fix: dumped buffer must not be empty

zardoy:
I already made this test in prismarine-anvil-provider:

```js
const dumped = oldChunk.dump()
const lights = oldChunk.dumpLight()
const biomes = oldChunk.dumpBiomes()

const chunk = new Chunk()
chunk.load(dumped, chunk.getMask(), true, true) // always expects buffer data to read
chunk.loadBiomes(biomes)
chunk.loadParsedLight?.(lights.skyLight, lights.blockLight, chunk.skyLightMask, chunk.blockLightMask, chunk.emptySkyLightMask, chunk.emptyBlockLightMask)
```

which runs against this file: https://github.com/zardoy/prismarine-provider-anvil/blob/everything/test/fixtures/1.13.2/r.-1.-1.mca which appears to be the world of type the void

and checked that native server always sends the buffer filled with 0 in these cases

### DIFF
--- a/src/pc/1.13/ChunkColumn.js
+++ b/src/pc/1.13/ChunkColumn.js
@@ -199,6 +199,10 @@ module.exports = (Block, mcData) => {
         smartBuffer.writeInt32BE(biome)
       })
 
+      if (!smartBuffer.length) {
+        return Buffer.alloc(4096)
+      }
+
       return smartBuffer.toBuffer()
     }
 
--- a/src/pc/1.13/ChunkColumn.js
+++ b/src/pc/1.13/ChunkColumn.js
@@ -200,7 +200,7 @@ module.exports = (Block, mcData) => {
       })
 
       if (!smartBuffer.length) {
-        return Buffer.alloc(4096)
+        return Buffer.alloc(1024)
       }
 
       return smartBuffer.toBuffer()
--- a/src/pc/1.14/ChunkColumn.js
+++ b/src/pc/1.14/ChunkColumn.js
@@ -221,6 +221,10 @@ module.exports = (Block, mcData) => {
         smartBuffer.writeInt32BE(biome)
       })
 
+      if (!smartBuffer.length) {
+        return Buffer.alloc(4096)
+      }
+
       return smartBuffer.toBuffer()
     }
 
--- a/src/pc/1.14/ChunkColumn.js
+++ b/src/pc/1.14/ChunkColumn.js
@@ -222,7 +222,7 @@ module.exports = (Block, mcData) => {
       })
 
       if (!smartBuffer.length) {
-        return Buffer.alloc(4096)
+        return Buffer.alloc(1024)
       }
 
       return smartBuffer.toBuffer()
--- a/src/pc/1.15/ChunkColumn.js
+++ b/src/pc/1.15/ChunkColumn.js
@@ -211,7 +211,7 @@ module.exports = (Block, mcData) => {
         }
       })
       if (!smartBuffer.length) {
-        return Buffer.alloc(4096)
+        return Buffer.alloc(1024)
       }
       return smartBuffer.toBuffer()
     }
--- a/src/pc/1.15/ChunkColumn.js
+++ b/src/pc/1.15/ChunkColumn.js
@@ -210,6 +210,9 @@ module.exports = (Block, mcData) => {
           section.write(smartBuffer)
         }
       })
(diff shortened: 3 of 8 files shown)


src/pc/1.13/ChunkColumn.js
```diff
@@ -199,6 +199,10 @@ module.exports = (Block, mcData) => {
         smartBuffer.writeInt32BE(biome)
       })
 
+      if (!smartBuffer.length) {
+        return Buffer.alloc(4096)
```
extremeheat, line 203: What are these zeros representing? Do you have a link to minecraft code to explain?

zardoy: It's just a workoaround so the load method (which expects the buffer to be not empty) doesn't crash. I'm not sure why I used the size of 4096. I see the native sever uses the length of 1024.

extremeheat: There should not be any "workarounds" in stable API methods. The libs should adhere to whatever the vanilla behavior is. In this case "load" deserializes a vanilla chunk from the network and "dump" serializes something to be sent over the network.

zardoy: Yes, in this case dump works incorrectly then. Afaik even minecraft-protocol doesn't allow serialization of empty buffers. I understand there should be no workarounds, can you elaborate on the correct way to fix it?

rom1504: Allocating a buffer of the right size

zardoy: Ok! Will also try to add a test

zardoy: >Allocating a buffer of the right size

@rom1504 ok, got it fixed. should be good now

src/pc/1.13/ChunkColumn.js
```diff
@@ -199,6 +199,10 @@ module.exports = (Block, mcData) => {
         smartBuffer.writeInt32BE(biome)
       })
 
+      if (!smartBuffer.length) {
+        return Buffer.alloc(1024)
```
rom1504, line 203: why do we want to create empty buffers? if there is a reason, add a comment

src/pc/1.13/ChunkColumn.js
rom1504: > It's just a workoaround so the load method (which expects the buffer to be not empty) doesn't crash

which load method do you mean? if you mean vanilla one; then ok we should do that and leave a comment to explain this is to address a bug in the vanilla implementation
if it's our implementation of load, then we should fix the bug

rom1504: > and checked that native server always sends the buffer filled with 0 in these cases

this seems to suggest the mojang server; is that indeed what you mean? Do you know why they have this bug?

zardoy: I don't think it's a bug, most probably the vanilla server just tried to normalize the data so the client marks received chunk as empty (otherwise it would print waiting for chunks)

zardoy: And I was about our load method. Not sure what should be fixed in that method instead

rom1504:
https://github.com/PrismarineJS/prismarine-provider-anvil/pull/81 looks like this is working without this change

is this prismarine chunk pr truly necessary?

zardoy:
> [PrismarineJS/prismarine-provider-anvil#81](https://github.com/PrismarineJS/prismarine-provider-anvil/pull/81) looks like this is working without this change
> 
> is this prismarine chunk pr truly necessary?

I never said it is required for basic functionality for new versions, it is not related. Can you look at OP please

rom1504:
I still don't understand what this PR is trying to fix

Is there a bug in the vanilla client or server that makes this necessary?

zardoy:
> I still don't understand what this PR is trying to fix

Sorry this is old, but as I can see in the description of this PR it fixes loading back the empty chunks. There is an attached region file for testing (let me see if it can be added as a test here).
You should be able to spot this bug via flying squid when trying to load world in mineflayer with region file as attached above

rom1504:
this is stuck, closing

zardoy:
> What you should do is figure out how it's possible to enter the situation rather than trying to error correct there. Sections.length==0 condition should I think throw an error.

ok I will do, should I continue my work here?

rom1504:
If something is broken yes

PR #241, opened 2024-06-12
fix block entity set

zardoy:
It never worked correctly.
not adding a test here as it's supposed to be used prismarine-provider-anvil obviously

### DIFF
--- a/src/pc/common/CommonChunkColumn.js
+++ b/src/pc/common/CommonChunkColumn.js
@@ -63,7 +63,7 @@ class CommonChunkColumn {
 
   loadBlockEntities (entities) {
     for (const entity of entities) {
-      this.setBlockEntity({ x: entity.x.value >> 4, y: entity.y.value, z: entity.z.value >> 4 }, entity)
+      this.setBlockEntity({ x: entity.x.value, y: entity.y.value, z: entity.z.value }, entity)
     }
   }
 }
--- a/src/pc/common/CommonChunkColumn.js
+++ b/src/pc/common/CommonChunkColumn.js
@@ -53,7 +53,6 @@ class CommonChunkColumn {
   }
 
   setBlockEntity (pos, tag) {
-    // Note: `pos` is relative to the chunk, not the world, tag's XYZ is
     this.blockEntities[posKey(pos)] = tag
   }
 


zardoy:
> If it's an exposed API there should definitely be a test, even if it's used by anvil-provider

hm, but how do you suggest testing it without prismarine-provider-anvil? Should I just hardcode arguments in a function call in a test?

eg like
```js
column.loadBlockEntities([
  {
    GlowingText: {
      type: "byte",
      value: 0,
    },
    keepPacked: {
      type: "byte",
      value: 0,
    },
    Color: {
      type: "string",
      value: "black",
    },
    Text4: {
      type: "string",
      value: "{\"text\":\"\"}",
    },
    x: {
      type: "int",
      value: 6,
    },
    Text3: {
      type: "string",
      value: "{\"text\":\"\"}",
    },
    y: {
      type: "int",
      value: -63,
    },
    Text2: {
      type: "string",
      value: "{\"text\":\"\"}",
    },
    z: {
      type: "int",
      value: 136,
    },
    Text1: {
      type: "string",
      value: "{\"text\":\"te\"}",
    },
    id: {
      type: "string",
      value: "minecraft:sign",
    },
  },
])
```

src/pc/common/CommonChunkColumn.js
```diff
@@ -53,7 +53,6 @@ class CommonChunkColumn {
   }
 
   setBlockEntity (pos, tag) {
-    // Note: `pos` is relative to the chunk, not the world, tag's XYZ is
```
extremeheat, line 56: Why the removal of this comment? The blockEntities object is supposed to indeed hold entities relative to the current chunk. The chunk is not necessarily bounded to any world hence it's not supposed to hold global coordinates.

rom1504:
this change seems incorrect since coordinates should be relative. Probably the issue is not here

zardoy:
> this change seems incorrect since coordinates should be relative. Probably the issue is not here

yes, now I see, making `coordinates be relative` actually makes sense so they are linked to the chunk and not the actual world holding the chunk. I will make sure to fix it properly.

I also wonder should block entities also dumped (using the `dump` method)?

rom1504:
closing as agreed above this is not necessary

PR #263, opened 2024-10-18
1.21

Madlykeanu:


rom1504:
we need test chunks like https://github.com/PrismarineJS/prismarine-chunk/tree/master/test/1.20

rom1504:
but let's do in another pr

PR #265, opened 2024-10-29
add is equal to diff in depth 2 chunks

rom1504:
started to implement this

but actually this might be misguided in some cases: the internal not being equal does not necessarily mean the chunk doesn't contain the same thing


rom1504:
let's not do it

PR #267, opened 2025-01-02
pc1.18+: Fix world height defaults and light masks

extremeheat:
The world height changed in 1.18+ to be between -64 and +320. The value set to the ChunkColumn have to be aligned with the dimension codec or the game will sometimes crash the game and otherwise cause chunks to be misplaced, eg:

image

Since we don't have a way to store this data inside prismarine-registry yet I just fix by changing the defaults to be correct 1.18 defaults

Also fix for light masks


rom1504:
would be nice to have a regression test for this kind of stuff

PR #292, opened 2025-07-06
fix bedrock level chunk and sub chunk parsing.

mc-zuri:
- Replace manual bedrock-provider test data copying with automated script
- Update README.md to document the new `generate-bedrock-test-data.mjs` script
- Fixed subchunks issue when y < 0
- Fixed stream.writeVarInt method to support negative numbers
- Fixed block hash ID reading/writing (#276)

### DIFF
--- a/tools/generate-bedrock-test-data.mjs
+++ b/tools/generate-bedrock-test-data.mjs
@@ -1,0 +1,426 @@
+import { createServer } from 'net'
+import { startServerAndWait2 } from 'minecraft-bedrock-server'
+import { createClient } from 'bedrock-protocol'
+import PrismarineRegistry from 'prismarine-registry'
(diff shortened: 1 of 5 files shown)


tools/generate-bedrock-test-data.mjs
extremeheat, line 1: The maintenence burden of having these server tests/generators across many repos with interdependence of bedrock-protocol doesn't seem ideal

extremeheat: Well, we don't have to remove it, I was just hoping it could go somewhere else. If the bedrock-provider tests are no longer generating it then for now I guess you can leave it in this repo.

extremeheat: No, bedrock-provider and anvil-provider are both for reading worlds from disk; chunk parsing code belongs in prismarine-chunk

PR #295, opened 2025-07-13
1.21.5 continued

rom1504:
added a fix that cursor proposed

### DIFF
--- a/src/pc/common/PaletteContainer.js
+++ b/src/pc/common/PaletteContainer.js
@@ -29,8 +29,8 @@ class DirectPaletteContainer {
 
   readBuffer (smartBuffer, bitsPerValue) {
     const longs = this.noSizePrefix
-      ? varInt.read(smartBuffer)
-      : Math.ceil(constants.BLOCK_SECTION_VOLUME * bitsPerValue / 64)
+      ? Math.ceil(constants.BLOCK_SECTION_VOLUME * bitsPerValue / 64)
+      : varInt.read(smartBuffer)
     this.data.readBuffer(smartBuffer, longs * 2)
     return this
   }
@@ -109,8 +109,8 @@ class IndirectPaletteContainer {
 
   readBuffer (smartBuffer, bitsPerValue) {
     const longs = this.noSizePrefix
-      ? varInt.read(smartBuffer)
-      : Math.ceil(constants.BLOCK_SECTION_VOLUME * bitsPerValue / 64)
+      ? Math.ceil(constants.BLOCK_SECTION_VOLUME * bitsPerValue / 64)
+      : varInt.read(smartBuffer)
     this.data.readBuffer(smartBuffer, longs * 2)
     return this
   }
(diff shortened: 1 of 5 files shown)


src/pc/common/PaletteContainer.js
```diff
@@ -29,7 +29,7 @@ class DirectPaletteContainer {
 
   readBuffer (smartBuffer, bitsPerValue) {
     const longs = this.noSizePrefix
-      ? Math.ceil(constants.BLOCK_SECTION_VOLUME * bitsPerValue / 64)
+      ? Math.ceil(constants.BLOCK_SECTION_VOLUME / Math.floor(64 / bitsPerValue))
```
rom1504, line 32: @extremeheat cursor proposed this. that's what it said


Now I understand the issue! The problem is in the size computation in PaletteContainer.js. The current formula is wrong. Let me analyze:
BLOCK_SECTION_VOLUME = 4096 (16x16x16)
The current formula: Math.ceil(constants.BLOCK_SECTION_VOLUME * bitsPerValue / 64)
But looking at the BitArray constructor, the correct formula should be: Math.ceil(capacity / valuesPerLong) * 2 where valuesPerLong = Math.floor(64 / bitsPerValue)
The issue is that we're calculating the number of longs needed, but we need to calculate the number of Uint32Array elements (which is longs * 2).


It does make mineflayer tests pass this issue

rom1504: not sure I get it though so if you could say whether that makes sense it'd be helpful

rom1504: ah I think I get it actually... ok I'll assume this is right and continue from that point

extremeheat: The original assumes no padding and the new one expects padding bits. Basically one block can't be split across two integers so it looks right

src/pc/common/PaletteContainer.js
```diff
@@ -22,12 +23,15 @@ class DirectPaletteContainer {
...
-    varInt.write(smartBuffer, this.data.length())
+    if (!this.noSizePrefix) varInt.write(smartBuffer, this.data.length())
     this.data.writeBuffer(smartBuffer)
   }
 
-  readBuffer (smartBuffer) {
-    this.data.readBuffer(smartBuffer, varInt.read(smartBuffer) * 2)
+  readBuffer (smartBuffer, bitsPerValue) {
+    const longs = this.noSizePrefix
+      ? Math.ceil(this.data.capacity / Math.floor(64 / bitsPerValue))
```
extremeheat, line 32: Previously there was a bug here in that it used a constant instead of checking underlying storage capacity:
> The core issue is that your generic IndirectPaletteContainer and DirectPaletteContainer classes are hardcoded to calculate their size based on a block section's volume (16x16x16 = 4096), even when they are being used for a biome section (4x4x4 = 64).
> The solution is to make the readBuffer methods in DirectPaletteContainer and IndirectPaletteContainer use their own capacity instead of the hardcoded BLOCK_SECTION_VOLUME. This will correctly calculate 1 long for biomes and 64 longs for blocks (when bitsPerValue is 1).

Biomes are no longer per block, but per chunks of a chunk column

PR #332, opened 2026-09-06
Fix Bedrock 1.17.30+ zero-bit palette decoding

Pix3lPirat3:
Select the version-9 subchunk decoder for Bedrock 1.17.30 and newer legacy chunk columns so zero-bit runtime palettes decode at the correct packet boundary.

The fix and tests cover both 1.17.30 and 1.17.40.


Pix3lPirat3:
Closing as a duplicate of #335 - same zero-bit 1.17.30 decode fix, but this one was opened from the fork's `master` branch. #335 carries it as a single clean commit; #337 is rebased to sit on top of it with just the hashed-state work.

Issue #229, opened 2023-08-13
Incorrectly decoding after sections (bedrock)


Flonja:
When sub chunks have been read properly border blocks, biomes and (possible) block entities follow.
Though when a server only uses `LevelChunkPacket`s, it can still send 1.18-type biomes, (unless the `vanilla_version` field is `1.17.40` or lower.)

This edge-case can be reproduced on a PMMP4 server, (as this doesn't use the mode `-1` or `-2`.)

[From (prismarine-chunk/1.18/ChunkColumn L145)](https://github.com/PrismarineJS/prismarine-chunk/blob/master/src/bedrock/1.18/ChunkColumn.js#L145) to the incorrect code block: [prismarine-chunk/1.3/ChunkColumn (L189-L191)](https://github.com/PrismarineJS/prismarine-chunk/blob/master/src/bedrock/1.3/ChunkColumn.js#L189-L191).
What it should be could be found in the previous implementation: [bedrock-provider/1.3/ChunkColumn.ts (L352-L363)](https://github.com/PrismarineJS/bedrock-provider/blob/97d1c386e7a3a9fe10c39e2e989a3360966ad1e9/src/chunk/1.3/ChunkColumn.ts#L352-L363).

Flonja:
Can't remember if this got fixed or not

PR #335, opened 2026-09-19
Fix Bedrock 1.17.30+ zero-bit palette decoding

Pix3lPirat3:
Select the version-9 subchunk decoder for the 1.3 chunk column when the registry is 1.17.30 or newer, so all-air (zero-bit) runtime palettes stop consuming the legacy biome bytes that follow the sections. PR 231 fixed the same symptom for the 1.18 column that issue 229 reported first; this covers the 1.3 column the issue also points at, used by 1.17.30 and 1.17.40 servers. Tests cover both versions and the version-8 path before 1.17.30.

Related: #229


rom1504 (commented):

No blocking issue found in this bounded decoder change. The full suite passes (331 tests), including the pre-1.17.30 version-8 control and the 1.17.30/1.17.40 zero-bit palette/biome-boundary roundtrips; lint passes. I checked that selection remains in the chunk factory and the runtime decode path uses that selected section implementation. I did not run a live Bedrock server.

The current Git tree is identical to #332, so this assessment also covers that code; only one of these duplicate PRs needs to land.

Skills used: `prismarine-protocol-data-review` checked the version boundary and actual decode path; `prismarine-architecture-review` checked factory ownership; `prismarine-review` checked the current diff and existing discussion.

src/bedrock/1.3/chunk.js
```diff
@@ -1,16 +1,23 @@
 const { ChunkVersion } = require('../common/constants')
 const ChunkColumn = require('./ChunkColumn')
+const ModernSubChunk = require('../1.18/SubChunk')
```
extremeheat: It's odd to load from 1.18 in 1.3 when src/bedrock/1.18 already exists

But, since 1.18 SubChunk is actually 1.17.30+ SubChunk, it seems the version handling code is incorrect at the moment.

SubChunk => SubChunkV9 for clarity and perhaps move src/bedrock/1.18/SubChunk.js to src/bedrock/1.17.30/SubChunkV9.js makes sense

But otherwise fix wise, PR LGTM

mc-zuri:
This is an automated code review from Claude. I ran the tests locally; please double-check the findings before acting on them.

Thanks for the fix! I tested it locally.

Results

- This PR's own suite passes: 331 passing, 0 failing (mocha only, I didn't run lint). The new test targets the reported bug directly: on 1.17.30 and 1.17.40, a chunk made of all-air subchunks (zero-bit palettes) decodes, reads back as air, and re-encodes to the same bytes. It also checks that 1.17.10 still uses subchunk version 8.
- I also ran the tests and fixtures from #327 against this branch, without #327's src/ changes. I used prismarine-registry from PrismarineJS/prismarine-registry#57 (installed from its branch, not yet released), which #327 depends on for block hashes, and minecraft-data 3.117.0 (latest).
  - This branch: 25 passing, 12 failing.
  - master with the same setup: the same 12 tests fail, so this PR doesn't cause any regressions, and its 3 new tests pass.
  - The 12 failures are all problems that #327's source changes fix:
    - level_chunk with caching on 1.18.0 and 1.21.60: ERR_OUT_OF_RANGE in BiomeSection.read
    - subchunk with caching on 1.18.0 and 1.21.60: Cannot read properties of undefined (reading 'updated'), when the subchunk index is below 0
    - subchunk without caching on 1.21.60 with block_network_ids_are_hashes = true: the re-encoded payload doesn't match the original
    - compaction on 1.21.60 with hashes: SubChunk.create throws reading 'name', because addToPalette looks blocks up in registry.blockStates, which doesn't work once registry#57 turns state IDs into hashes
    - 4 Stream varint tests: negative values aren't encoded correctly
- For reference, #327 on its own branch with the same registry#57 + minecraft-data 3.117.0 setup passes everything: 326 passing, 0 failing.

Overall this looks good to me as a fix for the all-air subchunk bug on 1.17.30+, and it doesn't break anything.

PR #336, opened 2026-09-19
Load chunk implementations lazily

Pix3lPirat3:
The version map required every pc and bedrock chunk decoder at import time, so `require('prismarine-chunk')` pulled in ~200 modules even though a process only ever uses the one implementation matching its bot's version. Each entry is now a thunk that `loader()` resolves on first use and memoises.

Measured on this checkout:
- cold `require` drops from ~280 ms to ~3 ms (2 modules instead of ~200)
- only the edition/version actually in use is ever loaded, so a pc-only bot never loads the bedrock decoders and vice versa (lower resident module count and memory)

For a process that does decode chunks the per-version build cost is deferred to the first chunk rather than removed. No behaviour change; the full suite (328 tests) passes and `standard` is clean.


rom1504 (commented):

No blocking issue found. The cache stores the implementation factory and still calls it with each caller’s registry, so it does not accidentally share a registry-bound chunk constructor across sessions. The full suite passes (328 tests), and lint passes. This is a correctness and package-boundary review; I have not independently reproduced the timing figures in the description.

Skills used: `prismarine-architecture-review` checked factory caching versus registry identity and the existing exports; `prismarine-review` checked the current diff, surrounding callers and existing discussion.

PR #337, opened 2026-09-19
Resolve hashed state ids when writing a sub-chunk block

Pix3lPirat3:
`SubChunk.addToPalette` (bedrock 1.3) indexed `registry.blockStates[stateId]`, but with hashed runtime ids (the `blockHashes` feature) the state id is a 32-bit hash, not an index into that array — so the lookup was `undefined` and reading its `.name` threw. Any single-block update (a placed or broken block) crashed the consumer.

Look the id up in the hash-keyed `blocksByStateId` first, fall back to `blockStates` for non-hashed registries, and tolerate an unknown id instead of dereferencing undefined. Full test suite passes.

### DIFF
--- a/src/bedrock/1.3/ChunkColumn.js
+++ b/src/bedrock/1.3/ChunkColumn.js
@@ -181,7 +181,7 @@ class ChunkColumn13 extends CommonChunkColumn {
     this.sections = []
     for (let i = 0; i < sectionCount; i++) {
       // in 1.17.30+, chunk index is sent in payload
-      const section = new SubChunk(this.registry, this.Block, { y: i, subChunkVersion: this.subChunkVersion })
+      const section = new this.Section(this.registry, this.Block, { y: i, subChunkVersion: this.subChunkVersion })
       section.decode(StorageType.Runtime, stream)
       this.setSection(i, section)
     }
--- a/src/bedrock/1.3/chunk.js
+++ b/src/bedrock/1.3/chunk.js
@@ -1,16 +1,23 @@
 const { ChunkVersion } = require('../common/constants')
 const ChunkColumn = require('./ChunkColumn')
+const ModernSubChunk = require('../1.18/SubChunk')
 
 module.exports = (version) => {
   // Require once here to avoid requiring() on every new chunk instance
   const registry = version.blockRegistry || version
+  const usesZeroBitRuntimePalettes = registry.version['>=']('1.17.30')
   const Block = require('prismarine-block')(registry)
   const Biome = require('prismarine-biome')(registry)
   return class Chunk extends ChunkColumn {
     constructor (options) {
       super(options, registry, Block, Biome)
       this.chunkVersion = this.chunkVersion || ChunkVersion.v1_16_0
-      this.subChunkVersion = 8
+      if (usesZeroBitRuntimePalettes) {
+        this.Section = ModernSubChunk
+        this.subChunkVersion = 9
+      } else {
+        this.subChunkVersion = 8
+      }
     }
 
     static fromJson (str) {
--- a/test/bedrock-1.17.30.test.js
+++ b/test/bedrock-1.17.30.test.js
@@ -1,0 +1,33 @@
+/* eslint-env mocha */
+
+const assert = require('assert')
+const { Vec3 } = require('vec3')
+
+for (const version of ['1.17.30', '1.17.40']) {
+  describe(`Bedrock ${version} zero-bit runtime palettes`, () => {
+    it('decodes all-air subchunks without consuming the legacy biome data', async () => {
+      const registry = require('prismarine-registry')(`bedrock_${version}`)
+      registry.handleStartGame({ block_network_ids_are_hashes: false, itemstates: [] })
+      const Chunk = require('..')(registry)
+      const Stream = require('../src/bedrock/common/Stream')
+      const stream = new Stream()
+
+      for (let y = 0; y < 32; y++) {
+        stream.writeUInt8(9)
+        stream.writeUInt8(1)
+        stream.writeUInt8(y)
+        stream.writeUInt8(1)
+        stream.writeZigZagVarInt(registry.blocksByName.air.defaultState)
+      }
+      stream.writeBuffer(Buffer.alloc(256, 18))
+      stream.writeUInt8(0)
+      const payload = stream.getBuffer()
+
+      const column = new Chunk({ x: 0, z: 0 })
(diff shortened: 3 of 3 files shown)


src/bedrock/1.3/SubChunk.js
```diff
@@ -271,8 +271,12 @@ class SubChunk {
   }
 
   addToPalette (l, stateId, count = 0) {
-    const block = this.registry.blockStates[stateId]
-    this.palette[l].push({ stateId, name: block.name, states: block.states, count })
+    // With hashed runtime ids (the blockHashes feature) stateId is a 32-bit state hash, not an index into the
+    // blockStates array, so look it up in the hash-keyed blocksByStateId table first and fall back to the index-ordered
+    // array for non-hashed registries. Tolerate an unknown state id rather than dereferencing undefined: a single
+    // block update carrying an unmapped id must not crash the whole client.
+    const block = this.registry.blocksByStateId?.[stateId] ?? this.registry.blockStates[stateId]
```
rom1504: **Astra agent review — AI-generated, not manually written by the maintainer.**

`blocksByStateId` in the declared/released registry contains block definitions, not the raw per-state NBT records needed here. Selecting it first drops the `states` payload even for ordinary internal state IDs.

With the actual `bedrock_1.26.45` registry, create a SubChunk, set an x-axis oak log (state 1530), then encode/decode it using either `StorageType.LocalPersistence` or `NetworkPersistence`. This head stores `{ stateId: 1530, name: 'oak_log', states: undefined }` and reads back state **1529**, the y-axis log. The same probe against the base addToPalette preserves 1530. This silently changes saved/cached block properties; all 331 existing tests still pass.

Please preserve the raw per-state NBT from `registry.blockStates` and resolve wire hashes through the registry's runtime-ID mapping before treating them as internal state IDs. In this registry, the x-axis log hash is -229389608; `blocksByRuntimeId[hash].stateId` is 1530, while `blocksByStateId[hash]` is absent. A non-default-property persistence roundtrip plus a hashed-update regression would cover both contracts without accepting undefined state data as a successful lookup.

Skills used: `prismarine-protocol-data-review` distinguished runtime hashes, internal state IDs and stored NBT; `prismarine-world-render-review` traced the registry through chunk persistence; `prismarine-review` checked both persistence formats against the base and existing feedback.

rom1504 (commented):

The suite passes (331 tests), but an additional real-registry persistence probe exposes loss of block properties in the changed palette lookup. Inline details include a passing base control. Skills used: `prismarine-protocol-data-review`, `prismarine-world-render-review`, and `prismarine-review` traced the ID/state contract and verified the regression.

Pix3lPirat3:
Fixed. addToPalette now reads the raw per-state record from registry.blockStates (which carries the per-state `states` NBT), instead of blocksByStateId (block definitions without it) that dropped the properties. Wire hashes are resolved through blocksByRuntimeId to the internal state id first, then read from blockStates. Added a non-default-property persistence round-trip test (oak_log pillar_axis=x, state 1530) on both LocalPersistence and NetworkPersistence - it read back 1530 now, vs 1529 (y-axis default) on the old code. Full suite 333 passing.

Note: bedrock_1.26.45's registry doesn't expose blocksByRuntimeId, so I couldn't construct a meaningful hashed-update regression on it here; the runtime-id resolution is a safe passthrough until that table is present (it's what the 1.18 SubChunk already uses).

src/bedrock/1.3/SubChunk.js
```diff
@@ -271,8 +271,14 @@ class SubChunk {
...
-    const block = this.registry.blockStates[stateId]
-    this.palette[l].push({ stateId, name: block.name, states: block.states, count })
+    // With hashed runtime ids (the blockHashes feature) stateId is a 32-bit state hash, not an index into blockStates:
+    // resolve it through the runtime-id map to the internal state id first. Then read the RAW per-state record from
+    // blockStates - it carries the per-state `states` NBT (the block's properties). blocksByStateId holds block
+    // definitions without that payload, so using it would drop the properties on a save/load round-trip. Tolerate an
+    // unknown id rather than dereferencing undefined: one unmapped block update must not crash the whole client.
+    const internalId = this.registry.blocksByRuntimeId?.[stateId]?.stateId ?? stateId
+    const block = this.registry.blockStates?.[internalId]
+    this.palette[l].push({ stateId, name: block?.name, states: block?.states, count })
```
rom1504, line 281: **Astra agent review — AI-generated, not manually written by the maintainer.**

The lookup now resolves the NBT record, but the palette still stores the original hash as `stateId`. This is reachable from decoding, without passing an invalid ID to the public setter: `SubChunk118.decode(Runtime)` → zero-bit `loadPalettedBlocks` → `addToPalette`.

With the declared/released registry, call `handleStartGame({ block_network_ids_are_hashes: true, itemstates: [] })` first; this populates `blocksByRuntimeId`. A version-9 zero-bit palette containing the known oak-log hash **825916963** resolves to internal state **1529**, yet this line stores 825916963. `getBlockStateId()` then returns that hash and `getBlock().name` is empty instead of `oak_log`. The ordinary runtime-palette decoder stores the resolved internal ID.

Please normalize the model’s palette `stateId` consistently, retaining a wire ID separately if needed, and exercise the actual zero-bit decode path with an initialized hashed registry. The earlier x-axis NBT persistence repair does pass; this is the remaining ID contract after that repair.

Skills used: `prismarine-protocol-data-review` followed the real negotiated registry and both ID representations; `prismarine-world-render-review` checked the decoded block model; `prismarine-architecture-review` checked consistency with the other palette loader.

rom1504 (commented):

Rechecked `096ddeb`: the earlier non-default-property persistence regression is fixed in both LocalPersistence and NetworkPersistence, and all 333 tests plus lint pass. Activating the real registry and exercising the zero-bit wire decoder exposes a remaining hashed-ID/model mismatch described inline.

Skills used: `prismarine-protocol-data-review` distinguished negotiated runtime hashes from internal state IDs; `prismarine-world-render-review` traced decoding into model access and persistence; `prismarine-review` verified the earlier repair instead of repeating the old finding.

Pix3lPirat3:
Rebased off the zero-bit 1.17.30 decode fix (which lands via #335), so this PR is now just the two hashed-state commits: resolve hashed state ids in `addToPalette` and keep the raw per-state NBT so a palette round-trip preserves block properties. 330 tests pass. The unknown-hash air fallback in the decode path is the separate #339 (different file, no conflict with this).

